PR #8, opened 2023-07-07
Fix grpc port conflict

PietroPasotti:
promtail tries to use port 9095 for its grpc listen server, but tempo has already snatched it.
This PR sets a different grpc listen port for Tempo.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,7 @@ from tempo import Tempo
 logger = logging.getLogger(__name__)
 
 
-@trace_charm(tempo_endpoint="_tempo_otlp_grpc_endpoint", service_name="TempoCharm")
+@trace_charm(tempo_endpoint="_tempo_otlp_grpc_endpoint")
 class TempoCharm(CharmBase):
     """Charmed Operator for Tempo; a distributed tracing backend."""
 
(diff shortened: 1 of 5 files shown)


lucabello (approved):

dstathis (approved):

src/charm.py
```diff
@@ -23,7 +23,7 @@
 logger = logging.getLogger(__name__)
 
 
-@trace_charm(tempo_endpoint="_tempo_otlp_grpc_endpoint", service_name="TempoCharm")
+@trace_charm(tempo_endpoint="_tempo_otlp_grpc_endpoint")
```
sed-i, line 26: Is there anything other than 'charm'? I.e. should it be shortened to just `@trace(...)`?

PietroPasotti: didn't see this one until now, but yeah, there's `trace_function, trace_type`.
Idea is:


```python
@trace_function
def my_custom_fn()...

@trace_type
class MyObj...

@trace_charm(...)
 class MyCharm...
     def foo(self):
          my_custom_fn()  # we'll open a span corresponding with `my_custom_fn` being called
          obj = MyObj()  # each method call on this type will open a span
          obj.bar()
```

In fact, `trace_charm` is a specialization of `trace_type`.
I thought about having a `@trace` that automagically dispatches to one of the three based on the type (and subclass) of the first argument, but it felt out of scope for now

PR #32, opened 2023-07-26
Endpoint removed handling

PietroPasotti:
- adds an `endpoint_removed` event for charm usage
- no longer raises exceptions when databag contents are invalid
- public `is_ready` check for charm usage

### DIFF
--- a/lib/charms/tempo_k8s/v0/tracing.py
+++ b/lib/charms/tempo_k8s/v0/tracing.py
@@ -470,11 +470,11 @@ class TracingEndpointProvider(Object):
         """Notify the providers that there is new endpoint information available."""
         relation = event.relation
         if not self.is_ready(relation):
+            self.on.endpoint_removed.emit(relation)  # type: ignore
             return
 
         data = TracingRequirerAppData.load(relation.data[relation.app])
-        if data:
-            self.on.endpoint_changed.emit(relation, data.host, [i.dict() for i in data.ingesters])  # type: ignore
+        self.on.endpoint_changed.emit(relation, data.host, [i.dict() for i in data.ingesters])  # type: ignore
 
     def _on_tracing_relation_broken(self, event: RelationBrokenEvent):
         """Notify the providers that the endpoint is broken."""
--- a/lib/charms/tempo_k8s/v0/tracing.py
+++ b/lib/charms/tempo_k8s/v0/tracing.py
@@ -64,7 +64,13 @@ import logging
 from typing import TYPE_CHECKING, List, Literal, MutableMapping, Optional, Tuple, cast
 
 import pydantic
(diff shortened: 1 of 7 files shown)


lib/charms/tempo_k8s/v0/tracing.py
```diff
@@ -425,8 +436,20 @@ def __init__(
...
         self.framework.observe(events.relation_changed, self._on_tracing_relation_changed)
+        self.framework.observe(events.relation_broken, self._on_tracing_relation_broken)
+
+    @property
+    def relation(self) -> Relation:
+        """The tracing relation associated with this endpoint.
+
+        Assumes tracing has limit 1.
+        """
+        return self._charm.model.get_relation(self._relation_name)
```
dstathis, line 447: Should we really be assuming limit 1 in a library.

dstathis (approved):

lib/charms/tempo_k8s/v0/tracing.py
PietroPasotti: yeah maybe it's not actually a bad idea to allow sending to multiple tracing endpoints

PietroPasotti: so I think we should keep it limit 1 because:
- I don't think it is common to have a single trace source support sending traces to multiple backends. (Traefik doesn't).
- if you really want to do that, the way to do it is to put an OpenTelemetry Collector or grafana agent or something like that in the middle, that distributes the traces to multiple backends.

PietroPasotti: @simskij sounds reasonable?

lucabello (approved):

lib/charms/tempo_k8s/v0/tracing.py
PietroPasotti: @lucabello does this make sense to you too?

lucabello: I think the two models you mentioned are something we might want to apply elsewhere as well; I'd say let's have a small chat with the whole team and see what approach we'd rather take :)

simskij: I think putting the GAgent in the middle if you want multiple sinks makes sense.

simskij (approved):

lib/charms/tempo_k8s/v0/tracing.py
PietroPasotti: GAgent will also use this lib's provider to push traces to multiple remote COS lite deployments (including multiple tempos). 
So the requirer lib should not assume limit 1. Traefik needs to set limit 1.

PR #43, opened 2023-09-04
Charm tracing name revert

PietroPasotti:
swapped out charm_instrumentation with charm_tracing lib

I just published charm_instrumentation 0.8 adding a warning about this rename and that charm_instrumentation is from now on unmaintained.


sed-i (approved):
1. Since the key tool is `trace_charm`, how about naming the library the same?
2. Would this tool make more sense under `observability_libs` or `cosl`?

PietroPasotti:
> 1. Since the key tool is `trace_charm`, how about naming the library the same?
> 2. Would this tool make more sense under `observability_libs` or `cosl`?

I think observability_libs would make sense, especially if we plan to have something like a `log_charm` that forwards juju-log to loki, etc... Let's ask @simskij when he's done tanning
About the name I don't know, I'd be fine either way but I like the sound of 'from ....charm_tracing import trace_charm'. It's got something poetic to it.

PR #45, opened 2023-09-12
downgraded some error logs to info

PietroPasotti:
when no relation is present on tracing, the charm_tracing lib logs a line telling the user that the module is effectively disabled as no traces can be forwarded.
This log used to be error. Now, info.

To test:
```charmcraft pack
juju deploy ./tempo-k8s_ubuntu-22.04-amd64.charm --resource tempo-image=grafana/tempo:1.5.0 tempo
juju debug-log
```

### DIFF
--- a/lib/charms/tempo_k8s/v0/tracing.py
+++ b/lib/charms/tempo_k8s/v0/tracing.py
@@ -61,17 +61,7 @@ follows
 """  # noqa: W505
 import json
 import logging
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    Literal,
-    MutableMapping,
-    Optional,
-    Tuple,
-    cast,
-)
+from typing import TYPE_CHECKING, List, Literal, MutableMapping, Optional, Tuple, cast
 
 import pydantic
 from ops.charm import (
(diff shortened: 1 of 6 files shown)


lib/charms/tempo_k8s/v0/tracing.py
```diff
@@ -484,7 +484,7 @@ def is_ready(self, relation: Optional[Relation] = None):
         """Is this endpoint ready?"""
         relation = relation or self._relation
         if not relation:
-            logger.error(f"no relation on {self._relation_name}: tracing not ready")
+            logger.info(f"no relation on {self._relation_name !r}: tracing not ready")
```
sed-i, line 487: Since tracing is an optional relation, I wonder if this should be `debug`. Do we need this log line at all?

sed-i (approved):

lib/charms/tempo_k8s/v0/tracing.py
PietroPasotti: "log everything" 
cit. @simskij

sed-i: A charm can have many relations but we do not log every charm reinit a "not ready" message for every relation.
Shouldn't a positive log suffice? I.e. "relation added", "relation removed" instead of "no relation".

PietroPasotti: this is not for every relation though: it's only once if there is no relation. 
it may be logged multiple times if the charm checks tracing readiness more than once, we could avoid that. But I think it's useful debug info to know that the charm couldn't find a tracing relation and therefore tracing is disabled.

PR #46, opened 2023-09-12
[doc] replace `charm_instrumentation` with `charm_tracing`

sed-i:
Follow up on #43 with more renames (doc fix).


PietroPasotti (approved):
good catch! Thanks

PietroPasotti (approved):

PR #51, opened 2023-11-23
pydantic v2 support

PietroPasotti:
this PR adds pydantic v2 support

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -10,3 +10,8 @@ bases:
       - name: "ubuntu"
         channel: "22.04"
 
+parts:
+  charm:
+    charm-binary-python-packages:
+      - "pydantic>=2"
+      - "opentelemetry-exporter-otlp-proto-grpc==1.17.0"
--- a/lib/charms/tempo_k8s/v0/tracing.py
+++ b/lib/charms/tempo_k8s/v0/tracing.py
@@ -83,7 +83,7 @@ from ops.charm import (
 )
 from ops.framework import EventSource, Object
 from ops.model import ModelError, Relation
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
 
 # The unique Charmhub library identifier, never change it
 LIBID = "12977e9aa0b34367903d8afeb8c3d85d"
@@ -93,9 +93,9 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
 
-PYDEPS = ["pydantic<2.0"]
+PYDEPS = ["pydantic>=2"]
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


mmkay (approved):

PietroPasotti:
Come to think of it, given we're effectively breaking BC and possibly introducing incompatibilities if you try and use this together with another charm (lib) that requires pydantic<2, should we actually be bumping the major?
@simskij @mmkay

sed-i:
> Come to think of it, given we're effectively breaking BC and possibly introducing incompatibilities if you try and use this together with another charm (lib) that requires pydantic<2, should we actually be bumping the major? @simskij @mmkay

Yeah, this PR should probably have quite a few tandem PRs.

lib/charms/tempo_k8s/v0/tracing.py
```diff
@@ -121,37 +122,61 @@ class AmbiguousRelationUsageError(TracingError):
...
-            msg = f"failed to validate remote unit databag: {databag}"
+            msg = f"failed to validate databag: {databag}"
             logger.error(msg, exc_info=True)
             raise DataValidationError(msg) from e
 
-    def dump(self, databag: MutableMapping):
-        """Write the contents of this model to Juju databag."""
-        if self._NEST_UNDER:
-            databag[self._NEST_UNDER] = self.json()
+    def dump(self, databag: Optional[MutableMapping] = None, clear: bool = True):
```
sed-i, line 158: **Suggestion (non-blocking)**: decouple `dump` from `update`.

Currently, `dump` does two things:
1. dump
2. update (a mutable, "by ref")

I think we should decouple. Decoupling might also obviate _NEST_UNDER and `dump` itself.

Could tackle this in a separate issue.

charmcraft.yaml
```diff
@@ -10,3 +10,8 @@ bases:
       - name: "ubuntu"
         channel: "22.04"
 
+parts:
+  charm:
+    charm-binary-python-packages:
+      - "pydantic>=2"
+      - "opentelemetry-exporter-otlp-proto-grpc==1.17.0"
```
sed-i, line 17: Would be handy to document deps here:

```suggestion
      # lib/charms/tempo_k8s/v0/tracing.py
      - "pydantic>=2"
      
      - "opentelemetry-exporter-otlp-proto-grpc==1.17.0"
```

sed-i (approved):

lib/charms/tempo_k8s/v0/tracing.py
```diff
@@ -93,9 +93,9 @@ def __init__(self, *args):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
```
mmkay, line 96: As #52 got merged, you'll need to bump up the version by one more.

```suggestion
LIBPATCH = 9
```

PietroPasotti: major bumped instead

Issue #33, opened 2023-07-26
tempo should activate as little ingesters as necessary


PietroPasotti:
### Enhancement Proposal

Right now, Tempo enables all ingesters: otlp_grpc, otlp_http, jaeger, zipkin...
That is wasteful.

Tempo should only enable the ingesters that the tracing providers connected to it need.

This will require to adjust the `tracing` protocol:
right now it is a one-way street: Tempo tells the providers: these are the ingesters I support.

it should be updated to do:
- each provider (app) tells Tempo what protocols it needs.
- Tempo configures itself to activate the providers needed to satisfy all requests (the union) 
- Tempo replies to each app with the ingester url(s) for each requested protocol.

mmkay:
That's already done by #65 .

PR #55, opened 2023-12-13
Switch to otlp/http protocol in charm_tracing

### LINKED ISSUE #53: Consider alternatives to GRPC for charm_tracing
### Enhancement Proposal

ATM charm-tracing uses tempo's otlp_grpc ingester to submit traces to Tempo
we should consider a more lightweight non-grpc (e.g. http) alternative to:
- prevent future compatibility issues when the grpc model versions end up desyncing 
- prevent adding 100MB to every charm using charm_tracing.

mmkay:
## Issue
Inclusion of `charm_tracing` library caused a dependency on GRPC that, in various contexts, added between 10MB to 100+MB to the charm size using this library (ref: canonical/prometheus-k8s-operator#545 ). We considered other protocols to send traces and decided that OTLP/HTTP seems to be the most relevant in the current context.

Resolves #53.

## Solution
Replacement of `opentelemetry-exporter-otlp-proto-grpc` with `opentelemetry-exporter-otlp-proto-http` together with bumping up to the latest (1.21.0) version.

The main difference between exporters is that instead of certificate being passed as one of the parameters of `OTLPSpanExporter`, HTTP exporter asks for a path to a certificate file. That means that the library upgrade for charms that are using a certificate (traefik-k8s-operator) requires changing the annotation signature and providing `certificate_file` instead of the current `certificate` parameter.

## Context
One interesting side effect of the version upgrade is that `_get_tracer()` couldn't find the context var for tracer in its context when an external function was instrumented using `@trace`. There was another context var reference with the same name pointing at the same tracer object in its context, though, so I added a fallback trace retrieval.

## Testing Instructions
Integration tests set up tempo together with a tester charm that sends traces and confirms they were ingested correctly.

## Release Notes
Replacement of OTLP/GRPC protocol with OTLP/HTTP in traces export.


lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -15,14 +15,14 @@
...
-that returns an otlp grpc endpoint url. If you are using the `TracingEndpointProvider` as
+that returns an otlp http endpoint url. If you are using the `TracingEndpointProvider` as
 `self.tracing = TracingEndpointProvider(self)`, the implementation could be:
 
 ```
     @property
     def my_tracing_endpoint(self) -> Optional[str]:
         '''Tempo endpoint for charm tracing'''
-        return self.tracing.otlp_grpc_endpoint
+        return f"http://{self.tracing.otlp_http_endpoint}/v1/traces"
```
PietroPasotti, line 25: shouldn't this URLification be done by the tracing lib?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +127,32 @@ def get_current_span() -> Union[Span, None]:
     return cast(Span, span)
 
 
+def _get_tracer_from_context(ctx: Context) -> Optional[ContextVar]:
+    try:
+        return next(filter(lambda v: v is not None and v.name == "tracer", ctx.keys()))
```
PietroPasotti, line 132: I know I'm guilty of using this pattern a bit too often myself, but I get reminded almost weekly that it's more readable to use a list comprehension instead.
```suggestion
    tracers = [v for v in ctx if v is not None and v.name == "tracer"]
    if tracers:
        return tracers[0]
        
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +127,32 @@ def get_current_span() -> Union[Span, None]:
...
+    """Find tracer in context variable and as a fallback locate it in the full context.
+
+    We need the fallback as `tracer` context variable isn't under the same reference in the instrumented function.
+    """
     try:
         return tracer.get()
     except LookupError:
-        return None
+        try:
+            logger.debug("tracer was not found in config variable, looking up in context")
```
PietroPasotti, line 146:
```suggestion
            logger.debug("tracer was not found in context variable, looking up in context")
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +127,32 @@ def get_current_span() -> Union[Span, None]:
...
+    try:
+        return next(filter(lambda v: v is not None and v.name == "tracer", ctx.keys()))
+    except StopIteration:
+        return
+
+
 def _get_tracer() -> Optional[Tracer]:
+    """Find tracer in context variable and as a fallback locate it in the full context.
+
+    We need the fallback as `tracer` context variable isn't under the same reference in the instrumented function.
```
PietroPasotti, line 140: not sure I understand what this means.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -244,11 +266,10 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
         server_cert: Optional[str] = (
             _get_server_cert(server_cert_getter, self, charm) if server_cert_getter else None
         )
-        credentials = ChannelCredentials(server_cert) if server_cert else None
-        insecure = None if credentials else True
 
+        # OTLP HTTP exporter expects a cert file path instead of certificate itself that OTLP GRPC did
```
PietroPasotti, line 270: remove this comment. it's good to know for this review but it doesn't need to remain there forever

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -327,11 +348,11 @@ def trace_charm(
     >>>         self.tempo = TracingEndpointProvider(self)
     >>>
     >>>     @property
-    >>>     def tempo_otlp_grpc_endpoint(self) -> Optional[str]:
-    >>>         return self.tempo.otlp_grpc_endpoint
+    >>>     def tempo_otlp_http_endpoint(self) -> Optional[str]:
+    >>>         return self.tempo.otlp_http_endpoint
```
PietroPasotti, line 352: should we URLify here as well?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -327,11 +348,11 @@ def trace_charm(
...
     >>>
     :param server_cert: method or property on the charm type that returns an
-        optional tls certificate to be used when sending traces to a remote server.
+        optional tls certificate path to be used when sending traces to a remote server.
         If it returns None, an _insecure_ connection will be used.
     :param tracing_endpoint: name of a property on the charm type that returns an
         optional tempo url. If None, tracing will be effectively disabled. Else, traces will be
```
PietroPasotti, line 358:
```suggestion
        optional (fully resolvable) tempo url. If None, tracing will be effectively disabled. Else, traces will be
```

PietroPasotti (changes requested):
Looks great! A couple of concerns/API design questions we can discuss about

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Seems that comment wasn't relevant anymore. Removed the second line.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Good point, fixed.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +127,31 @@ def get_current_span() -> Union[Span, None]:
     return cast(Span, span)
 
 
+def _get_tracer_from_context(ctx: Context) -> Optional[ContextVar]:
+    try:
+        tracers = [v for v in ctx if v is not None and v.name == "tracer"]
+        if tracers:
+            return tracers[0]
+    except StopIteration:
+        return
```
Abuelodelanada, line 136: Does `ctx` raise a `StopIteration` exception? 🤔

mmkay: Right, that was a leftover from previous approach. Removed the catch now.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: I changed the approach as discussed. Now:
- `tracing` lib adds the protocol part if it's needed based on whether we're using one of the HTTP protocols and SSL is enabled or not
- `charm_tracing` lib, as it uses `otlp/http`, adds the `/v1/traces` suffix to the URLs so `tracing` lib can stay generic for other instrumentation libraries

There's a potential of further improvements to remove the need to specify the endpoint altogether in future but I'll create a separate enhancement issue for that. 

Please have a look.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Made both use the same approach.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -15,14 +15,17 @@
 `@trace_charm(tracing_endpoint="my_tracing_endpoint")`
 
 2) add to your charm a "my_tracing_endpoint" (you can name this attribute whatever you like) **property**
-that returns an otlp grpc endpoint url. If you are using the `TracingEndpointProvider` as
+that returns an otlp http endpoint url. If you are using the `TracingEndpointProvider` as
```
PietroPasotti, line 18: perhaps worth clarifying here that it can be a http or https url.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +131,29 @@ def get_current_span() -> Union[Span, None]:
...
+    return None
+
+
 def _get_tracer() -> Optional[Tracer]:
+    """Find tracer in context variable and as a fallback locate it in the full context."""
     try:
         return tracer.get()
     except LookupError:
-        return None
+        try:
```
PietroPasotti, line 146: Why did you add this, and what is the use case? I don't remember having trouble using _get_tracer

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -141,6 +163,7 @@ def _span(name: str) -> Generator[Optional[Span], Any, Any]:
         with tracer.start_as_current_span(name) as span:
             yield cast(Span, span)
     else:
+        logger.debug(f"no tracer found: {_get_tracer()}")
```
PietroPasotti, line 166:
```suggestion
        logger.debug(f"tracer not found")
```

it's probably going to be None in that case, no?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -192,7 +215,7 @@ def _get_server_cert(server_cert_getter, self, charm):
         return
     elif not isinstance(server_cert, str):
         raise TypeError(
-            f"{charm}.{server_cert_getter} should return a valid tls cert (string); "
+            f"{charm}.{server_cert_getter} should return a valid tls cert path (string); "
```
PietroPasotti, line 218:
```suggestion
            f"{charm}.{server_cert_getter} should return the absolute path to a tls cert file"
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -192,7 +215,7 @@ def _get_server_cert(server_cert_getter, self, charm):
         return
     elif not isinstance(server_cert, str):
```
PietroPasotti, line 216: should probably take `str | Path` and make it clear we're assuming it's absolute

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -244,11 +267,9 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
         server_cert: Optional[str] = (
             _get_server_cert(server_cert_getter, self, charm) if server_cert_getter else None
         )
-        credentials = ChannelCredentials(server_cert) if server_cert else None
-        insecure = None if credentials else True
 
         exporter = OTLPSpanExporter(
-            endpoint=tracing_endpoint, credentials=credentials, insecure=insecure, timeout=2
+            endpoint=tracing_endpoint, certificate_file=server_cert, timeout=2
```
PietroPasotti, line 272:
```suggestion
            endpoint=tracing_endpoint, certificate_file=str(Path(server_cert).abspath()), timeout=2
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -313,28 +334,31 @@ def trace_charm(
...
-    >>>         return self.tempo.otlp_grpc_endpoint
+    >>>     def tempo_otlp_http_endpoint(self) -> Optional[str]:
+    >>>         if self.tracing.is_ready():
+    >>>             return self.tracing.otlp_http_endpoint()
+    >>>         else:
+    >>>             return None
     >>>
     :param server_cert: method or property on the charm type that returns an
-        optional tls certificate to be used when sending traces to a remote server.
+        optional tls certificate path to be used when sending traces to a remote server.
```
PietroPasotti, line 358:
```suggestion
        optional absolute path to a tls certificate to be used when sending traces to a remote server.
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -370,19 +394,20 @@ def _autoinstrument(
...
+    >>>         tracing_endpoint_getter=MyCharm.tempo_otlp_http_endpoint,
     >>>         service_name="MyCharm",
     >>>         extra_types=(Foo, Bar)
     >>> )
     >>> main(MyCharm)
 
     :param charm_type: the CharmBase subclass to autoinstrument.
     :param server_cert_getter: method or property on the charm type that returns an
-        optional tls certificate to be used when sending traces to a remote server.
+        optional tls certificate path to be used when sending traces to a remote server. This
```
PietroPasotti, line 409:
```suggestion
        optional absolute path to a tls certificate to be used when sending traces to a remote server.
        ```

File: lib/charms/tempo_k8s/v1/charm_tracing.py [hunk 17]
```diff
@@ -370,19 +394,20 @@ def _autoinstrument(
...
+        needs to be a valid path to a certificate.
```
PietroPasotti, line 410:
```suggestion
        needs to be a valid path to a certificate.
```
```suggestion
```

lib/charms/tempo_k8s/v1/tracing.py
```diff
@@ -547,37 +537,65 @@ def get_all_endpoints(
...
-    def _get_ingester(self, relation: Optional[Relation], protocol: IngesterProtocol):
+    def _get_ingester(self, relation: Optional[Relation], protocol: IngesterProtocol, ssl: bool):
         ep = self.get_all_endpoints(relation)
         if not ep:
             return None
         try:
             ingester: Ingester = next(filter(lambda i: i.protocol == protocol, ep.ingesters))
-            return f"{ep.host}:{ingester.port}"
+            if ingester.protocol in ["otlp_grpc", "jaeger_grpc"]:
+                return f"{ep.host}:{ingester.port}"
```
PietroPasotti, line 547: pop a warning "unused argument" if ssl=True?

lib/charms/tempo_k8s/v1/tracing.py
```diff
@@ -547,37 +537,65 @@ def get_all_endpoints(
...
-    def jaeger_http_thrift_endpoint(self, relation: Optional[Relation] = None) -> Optional[str]:
-        """Ingester endpoint for the ``jaeger_http_thrift`` protocol."""
-        return self._get_ingester(relation or self._relation, "jaeger_http_thrift")
+        The provided endpoint does not contain the endpoint suffix. If the instrumenting library needs the full path,
+        your endpoint code needs to add the ``/api/traces`` suffix.
+        """
+        return self._get_ingester(relation or self._relation, "jaeger_http_thrift", ssl=ssl)
 
-    def jaeger_grpc_endpoint(self, relation: Optional[Relation] = None) -> Optional[str]:
+    def jaeger_grpc_endpoint(
```
PietroPasotti, line 597: does ssl make sense with grpc? we shouldn't add it as an arg to this method if we don't plan on doing anything with it

PietroPasotti: (same with the other grpc protocol)

src/charm.py
```diff
@@ -136,11 +135,11 @@ def _get_version(self) -> Optional[str]:
             return
         return version
 
-    def tempo_otlp_grpc_endpoint(self) -> Optional[str]:
+    def tempo_otlp_http_endpoint(self) -> Optional[str]:
         """Endpoint at which the charm tracing information will be forwarded."""
         # the charm container and the tempo workload container have apparently the same
         # IP, so we can talk to tempo at localhost.
-        return f"http://localhost:{self.tempo.otlp_grpc_port}/"
+        return f"http://localhost:{self.tempo.otlp_http_port}/v1/traces"
```
PietroPasotti, line 142: should we be using https here btw? I don't think we have tls on Tempo yet, perhaps we should add a TODO here for when that's done?

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: This is quite interesting. After the switch to the new dependency it seems there's been a change in the way `tracer` behaves in the context here. I started getting scenario test failures on `test_base_tracer_endpoint_event` so I started debugging and noticed that `_get_tracer()` returns None in the context of the span in the function instrumented by `@trace`. After going through the context I noticed that there is a `tracer` in the context but with a different reference, so I added this fallback retrieval.

To test it doesn't only happen on my local environment I've tried removing the fallback on a test draft PR and it looks the same behavior happens there: https://github.com/canonical/tempo-k8s-operator/actions/runs/7412918066/job/20170743398?pr=58

Branch diff: https://github.com/canonical/tempo-k8s-operator/compare/switch-to-otlp-http...temp/otlp-http-without-get-context

As fallback is triggered only in case `tracer.get()` doesn't find anything I felt it's okay to add it here in that way. I must admit though that it's dealing with a symptom, not addressing the cause I couldn't really pinpoint.

Do you think we should open a bug to investigate that further?

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: True. Updated the comment here.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Correct. Looks much better, changed.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Done.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Done with a tiny difference as `Path` has `.absolute()` function that does the same.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Updated, thanks!

lib/charms/tempo_k8s/v1/tracing.py
mmkay: Makes sense, done.

lib/charms/tempo_k8s/v1/tracing.py
mmkay: On second thought the path GRPC dependencies use is the same regardless if SSL is used or not so I removed the parameter.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: True, took that into account and added a check if the path is absolute.

src/charm.py
mmkay: Added a TODO in here.

mmkay:
@PietroPasotti 
> I'm assuming you tried it out manually to see if it works?

Correct - manual reproduction steps, assuming tempo is mounted on the multipass host under `~/tempo-k8s-operator` and libraries are copied to tester charm:

```
juju add-model cos
juju switch cos
juju deploy cos-lite --trust
cd tempo-k8s-operator
charmcraft pack 
charmcraft pack -p ./tests/integration/tester/
juju deploy ./tempo-k8s_ubuntu-22.04-amd64.charm tempo --resource tempo-image=grafana/tempo:1.5.0
juju deploy ./tester_ubuntu-22.04-amd64.charm tester -n 3 --resource workload=python:slim-buster
juju relate tempo:tracing tester:tracing
juju relate grafana:grafana-source tempo:grafana-source
```

You should be able to see traces from TempoTesterCharm in Grafana:
Screenshot from 2024-01-04 20-57-55

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -62,33 +66,34 @@ def tracer(self) -> opentelemetry.trace.Tracer:
...
+LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
 
-PYDEPS = ["opentelemetry-exporter-otlp-proto-grpc==1.17.0"]
+LIBPATCH = 0
+
+PYDEPS = ["opentelemetry-exporter-otlp-proto-http==1.21.0"]
```
sed-i, line 96: Can we loosen the pin to `>=`?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +132,29 @@ def get_current_span() -> Union[Span, None]:
     return cast(Span, span)
 
 
+def _get_tracer_from_context(ctx: Context) -> Optional[ContextVar]:
+    tracers = [v for v in ctx if v is not None and v.name == "tracer"]
```
sed-i, line 136: Just curious:
- Where is the `"tracer"` literal coming from?
- How could we end up with multiple contextvars that have the same name?
- Are we assuming here that different "tracer" contextvars would have the same value? I.e. why do we return `tracersp[0]` only?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +132,29 @@ def get_current_span() -> Union[Span, None]:
...
+    if tracers:
+        return tracers[0]
+    return None
+
+
 def _get_tracer() -> Optional[Tracer]:
+    """Find tracer in context variable and as a fallback locate it in the full context."""
     try:
         return tracer.get()
```
sed-i, line 145: At first glance, `tracer` seems undefined. Where is it coming from?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -127,11 +132,29 @@ def get_current_span() -> Union[Span, None]:
...
     except LookupError:
-        return None
+        try:
+            logger.debug("tracer was not found in context variable, looking up in context")
```
sed-i, line 148: Not sure, but should it be:
```suggestion
            logger.debug("tracer was not found in context variable, looking up in default context")
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -141,6 +164,7 @@ def _span(name: str) -> Generator[Optional[Span], Any, Any]:
         with tracer.start_as_current_span(name) as span:
             yield cast(Span, span)
     else:
+        logger.debug("tracer not found")
```
sed-i, line 167: Leftover?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -190,9 +214,9 @@ def _get_server_cert(server_cert_getter, self, charm):
             f"{charm}.{server_cert_getter} returned None; continuing with INSECURE connection."
         )
         return
-    elif not isinstance(server_cert, str):
+    elif not (isinstance(server_cert, (str, Path)) and Path(server_cert).is_absolute()):
```
sed-i, line 217: Do we need the instance check? How about:
```suggestion
    elif not Path(server_cert).is_absolute():
```

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -190,9 +214,9 @@ def _get_server_cert(server_cert_getter, self, charm):
...
         raise TypeError(
```
sed-i, line 218: At this point, if we use just Path above, this would no longer be a TypeError.

tests/integration/tester/src/charm.py
```diff
@@ -98,7 +98,7 @@ def _tester_layer(self):
             "PORT": self.config["port"],
             "HOST": self.config["host"],
             "APP_NAME": self.app.name,
-            "TEMPO_ENDPOINT": str(self.tracing.otlp_grpc_endpoint) or "",
+            "TEMPO_ENDPOINT": str(self.tracing.otlp_http_endpoint) or "",
```
sed-i, line 101:
```suggestion
            "TEMPO_ENDPOINT": str(self.tracing.otlp_http_endpoint or ""),
```
But should it be / is it necessary for the user to convert a `None` to a `""`?

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: yeah something doesn't feel right. Maybe the pydantic warnings are related? We bumped to >2, maybe that broke something somewhere? No need to fix it in this PR but let's open an issue to remind ourselves

mmkay: I see you created #60 for that, thanks!

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: it's a global

tests/integration/tester/src/charm.py
PietroPasotti: besides, `str(None)` will be `"None"` and therefore truthy, so that's a bug either way :)

charmcraft.yaml
```diff
@@ -14,4 +14,4 @@ parts:
   charm:
     charm-binary-python-packages:
       - "pydantic>=2"
-      - "opentelemetry-exporter-otlp-proto-grpc==1.17.0"
```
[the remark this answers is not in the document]
PietroPasotti: whether this dependency lives here or in requirements.txt, charmcraft will pick it up regardless and packing time will be the same.

(we usually put it in both though, otherwise tox will not automatically recognize it as a dependency and we need to manually add it to all envs)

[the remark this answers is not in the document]
PietroPasotti: the change is compatible with requirer charms that use grpc, but it will require them to ALSO install the otlp_http lib.

ATM a requirer charm obtains from tempo all endpoints (not just the ones it needs: see #33 )

So the upgrade story for  charm_tracing v0 --> v1 is going to be:

(for tracing-requirer charms)
- bump charm lib to v1
- replace the grpc endpoint url to the otlp_http one in the charm method `@trace` is pointing to (using the existing `TracingRequirer.otlp_http_endpoint()` method)
- instead of returning a ssl cert, return a path to it from the other method (if using tls)
- profit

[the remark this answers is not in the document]
PietroPasotti: so if a charm was using otlp_grpc to send traces (and to no other purpose), they can drop the python dependency and add one to otlp_http instead.

If they also need otlp_grpc for their own thing, that's not our problem.

Just like before this change, if a charm depended on otlp_http for their own business, we were forcing them to also depend on otlp_grpc.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Tracer was defined here: https://github.com/canonical/tempo-k8s-operator/blob/switch-to-otlp-http/lib/charms/tempo_k8s/v1/charm_tracing.py#L100

The fallback code comes from the (unexplained yet) realization that the same tracer is available under a different reference in the context of a function decorated with `@trace`. We created #60 to investigate further.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Not really: that was on purpose as I spent a day trying to find out that it was the missing tracer contextvar. So in case we find one more bug this line might hopefully lead the poor investigator to a conclusion sooner.

tests/integration/tester/src/charm.py
mmkay: Nice catch! Fixed.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: We likely don't anymore, removed the instance check.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Replaced it with ValueError - seems more valid.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Sounds better, updated.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: I think we can, so I changed it to `>=`.

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: is it me, or I don't see the update?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -15,14 +15,17 @@
...
 ```
     @property
     def my_tracing_endpoint(self) -> Optional[str]:
         '''Tempo endpoint for charm tracing'''
-        return self.tracing.otlp_grpc_endpoint
+        if self.tracing.is_ready():
+            return self.tracing.otlp_http_endpoint()
+        else:
+            return None
 ```
```
PietroPasotti, line 29: It's not standard practice (yet, but why not), but I think it would be a good idea to add in this docstring an "upgrading from v0" section giving a minimal example of what you need to change in charm code to port a charm to use the new lib.
No need to mention charmcraft fetch-lib, just focus on 'user code'.

Not a hard requirement for this PR, just a thought if you like the idea. Otherwise we can document it elsewhere.

PietroPasotti (approved):
Noice! GJ. No need from another pass from my side.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Nice idea, I added a short `upgrading from v0` section at the end of this docstring.

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Seems like this comment landed on a subset of commits: when you pick all commits, it's there: https://github.com/canonical/tempo-k8s-operator/pull/55/files#diff-7ad9005cca2d3157d4ec395255a4a9ec78900fb4a91288d532c24cd64fc74176R18

lib/charms/tempo_k8s/v1/charm_tracing.py
sed-i: Is "tracer not found" a bit too ambiguous?
I wonder if something like this could be useful:
```python
>>> def f():
...     print(" -> ".join(map(lambda v: v.function, inspect.stack())))
... 
>>> # also define g and h
>>> h()
f -> g -> h -> <module>
```

PR #67, opened 2024-02-06
Reduce log spam on looking up tracer

### LINKED ISSUE #66: Log spam from the trace lib
### Enhancement Proposal

When there are no tracing relation in place, we still frequently get repetitive log lines:
```
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log Emitting Juju event update_status.
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log tracer was not found in context variable, looking up in default context
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log Couldn't find context var for tracer: span will be skipped
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log tracer not found
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log tracer was not found in context variable, looking up in default context
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log Couldn't find context var for tracer: span will be skipped
unit-prometheus-0: 19:22:18 DEBUG unit.prometheus/0.juju-log tracer not found
```

This happens every hook so it would be nice not to see it when it's not relevant.

mmkay:
## Issue
Reduce log spam on tracer context variable lookup. These logs aren't relevant in most cases.

Fixes #66

### DIFF
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -146,7 +146,7 @@ LIBAPI = 1
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 1
+LIBPATCH = 2
 
 PYDEPS = ["opentelemetry-exporter-otlp-proto-http>=1.21.0"]
 
@@ -200,15 +200,12 @@ def _get_tracer() -> Optional[Tracer]:
         return tracer.get()
     except LookupError:
         try:
-            logger.debug("tracer was not found in context variable, looking up in default context")
             ctx: Context = copy_context()
             if context_tracer := _get_tracer_from_context(ctx):
                 return context_tracer.get()
             else:
-                logger.debug("Couldn't find context var for tracer: span will be skipped")
                 return None
         except LookupError as err:
-            logger.debug(f"Couldn't find tracer: span will be skipped, err: {err}")
             return None
 
 
@@ -219,7 +216,6 @@ def _span(name: str) -> Generator[Optional[Span], Any, Any]:
         with tracer.start_as_current_span(name) as span:
             yield cast(Span, span)
     else:
-        logger.debug("tracer not found")
         yield None
 
 
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -240,8 +240,8 @@ def _get_tracing_endpoint(tracing_endpoint_getter, self, charm):
 
     if tracing_endpoint is None:
         logger.debug(
-            f"{charm}.{getattr(tracing_endpoint_getter, '__qualname__', str(tracing_endpoint_getter))} "
-            f"returned None; continuing with tracing DISABLED."
+            "Charm tracing is disabled. Tracing endpoint is not defined - "
+            "tracing is not available or relation is not set."
         )
         return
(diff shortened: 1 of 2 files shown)


sed-i (commented):
Nice, thanks!
How about:
- "Initializing opentelemetry"
- "tracing not ready"
- "span will be skipped"

mmkay:
@sed-i : some of the logs you mentioned are in the context of `tracing` lib and point exactly at the relation not being present and don't make sense once the relation is established. In `charm_tracing` we don't have the notion of the relation altogether. I agree, though, that some of them don't add value / shouldn't be warnings. I did one more prune in `charm_tracing` so that log overhead is reduced.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -265,7 +261,7 @@ def _get_server_cert(server_cert_getter, self, charm):
         server_cert = server_cert_getter(self)
 
     if server_cert is None:
-        logger.warning(
+        logger.debug(
             f"{charm}.{server_cert_getter} returned None; continuing with INSECURE connection."
         )
```
sed-i, line 266: Is this logline needed?

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -243,7 +239,7 @@ def _get_tracing_endpoint(tracing_endpoint_getter, self, charm):
         tracing_endpoint = tracing_endpoint_getter(self)
 
     if tracing_endpoint is None:
-        logger.warning(
+        logger.debug(
             f"{charm}.{getattr(tracing_endpoint_getter, '__qualname__', str(tracing_endpoint_getter))} "
             f"returned None; continuing with tracing DISABLED."
```
sed-i, line 244: This sounds like either normal operation and should log, or misuse of the lib and should raise an exception.

sed-i (approved):

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: It is a normal operation in case tracing endpoint is not defined, for instance because tracing relation isn't set yet. I feel though that if the charm has charm_tracing support it might be something the user might want to know. Again, though, I think we could have a better message.
```suggestion
        logger.debug(
            "Charm tracing is disabled. Tracing endpoint is not defined - tracing is not available or relation is not set."
```

PietroPasotti: +1

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: if we are given a cert getter, but the cert getter returns no cert, I think we should log something perhaps even at warning, because we might be transferring sensitive data in the clear, for all we know

PietroPasotti: if we are not given a cert getter, then we assume the owner is fine with insecure connections

mmkay: Indeed we go into this flow only if cert getter is defined. We might want to stay at warning here then:
```suggestion
        logger.warning(
            f"{charm}.{server_cert_getter} returned None; sending traces over INSECURE connection."
        )
```

PR #73, opened 2024-02-19
Check if tracing is ready before exposing its endpoint in tempo charm

mmkay:
## Issue
There's a transient error status on tempo startup as charm-tracing lib races with tempo startup.

## Solution
Add an `is_ready` guard before exposing tempo endpoint.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -263,7 +263,7 @@ class TempoCharm(CharmBase):
         # the charm container and the tempo workload container have apparently the same
         # IP, so we can talk to tempo at localhost.
         # TODO switch to HTTPS once SSL support is added
-        if self._tracing.is_ready():
+        if self.tempo.is_ready():
             return f"http://localhost:{self.tempo.receiver_ports['otlp_http']}"
         else:
             return None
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -263,7 +263,10 @@ def tempo_otlp_http_endpoint(self) -> Optional[str]:
         # the charm container and the tempo workload container have apparently the same
         # IP, so we can talk to tempo at localhost.
         # TODO switch to HTTPS once SSL support is added
-        return f"http://localhost:{self.tempo.receiver_ports['otlp_http']}"
+        if self.tempo.is_ready():
+            return f"http://localhost:{self.tempo.receiver_ports['otlp_http']}"
+        else:
+            return None
```
Abuelodelanada, line 269: Just a matter of taste, the `else` it's not necessary 

```suggestion
        if self.tempo.is_ready():
            return f"http://localhost:{self.tempo.receiver_ports['otlp_http']}"

        return None
```

Abuelodelanada (approved):

PR #79, opened 2024-03-13
[WIP] snapshot lib POC

PietroPasotti:
This PR adds a charm library that offers a charm the possibility to snapshot its own state.
State here refers to scenario's `State` data structure.

Internally it uses a ripoff of the `jhack scenario snapshot` functionality.

Goal is to offer charms the possibility to add state snapshots to traces as event datapoints for troubleshooting.

This PR also updates charm_tracing.py to expose API to hook up snapshotting automatically from the `@trace_charm` decorator as one would expect:

```python
from charms.tempo_k8s.v0.snapshot import get_state
@trace_charm(
        tracing_endpoint="something",
        state="state",
)
class MyCharm(CharmBase):
    @property
    def state(self) -> scenario.State:
        return get_state()
```

(that will be split out to a separate PR once the design is validated)

Result: `scenario.State`s attached to the `"charm_exec"` root span in grafana!

image


lib/charms/tempo_k8s/v0/snapshot.py
```diff
@@ -0,0 +1,187 @@
...
+    def get_app_status(self) -> ops.StatusBase:  # noqa: D102
+        if not self.charm.unit.is_leader():
+            return ops.UnknownStatus()
+        return self.charm.app.status
+
+    def get_unit_status(self) -> ops.StatusBase:  # noqa: D102
+        return self.charm.unit.status
+
+    def get_workload_version(self) -> str:  # noqa: D102
+        return "42.todo"
```
mmkay, line 68: That one likely can disappear already

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -413,6 +508,8 @@ def trace_charm(
     :param tracing_endpoint: name of a property on the charm type that returns an
         optional (fully resolvable) tempo url. If None, tracing will be effectively disabled. Else, traces will be
         pushed to that endpoint.
+    :param tracing_endpoint: name of a property on the charm type that returns an
```
mmkay, line 511: `tracing_endpoint` -> `state`

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -387,12 +475,15 @@ def trace_charm(
     method calls on instances of this class.
 
     Usage:
+    >>> import scenario
```
mmkay, line 478: I wonder if this should be part of the default path since `state` is optional. Will `charm_tracing` work without importing scenario and snapshot lib? If so, we might want to split the documentation into two versions - basic as a minimal instrumentation and extended with `state` and `extra_types`.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -560,3 +660,111 @@ def trace(obj: Union[Type, Callable]):
...
+
+def _dict_to_deferred(value: Dict) -> DeferredEvent:
+    return DeferredEvent(**value)
+
+
+def _dict_to_storage(value: Dict) -> Storage:
+    return Storage(**value)
+
+
+def dict_to_state(state_json: Dict) -> State:  # noqa: C901
```
mmkay, line 730: I do see why you want to put conversions both ways (both `state -> dict` and `dict -> state`) close to the place where it's used but I wonder if that's part of `charm_tracing` area of responsibility.

Especially this method (in the dict -> state direction), together with the helper methods above, is currently essentially used within the tests? I'm thinking if it might be a better fit within the `snapshot` lib. Or a converter class that _could_ be outside of charm libs altogether, maybe closer to scenario? I can imagine this being useful in other contexts than `charm_tracing` as well. Please correct me if I missed something or don't yet see where that's heading.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -147,7 +173,7 @@ def my_tracing_endpoint(self) -> Optional[str]:
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 4
+LIBPATCH = 5
 
 PYDEPS = ["opentelemetry-exporter-otlp-proto-http>=1.21.0"]
```
mmkay, line 178: If we import scenario, we also start being dependent on scenario here, right?

mmkay (changes requested):
Nice work here, I really like the shown result! I have some doubts around the location of code converting between dicts to be saved in traces and scenario state. Let's discuss if we can figure something out for this part.

lib/charms/tempo_k8s/v0/snapshot.py
PietroPasotti: yeah, turns out a charm can't access its own workload version

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: good idea. Will put that in the module docstring instead.

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: you're right, it's ugly. 

IMHO: 
- the snapshot lib is responsible for generating a State object
- the charm (user code) shouldn't be responsible for serializing it to something that the charm_tracing lib can consume. 
- charm_tracing needs 

So either we make the get_state function return an opaque serialized object that the charm user can't inspect (might be fine if we want to prevent misuse, but misuse is in this case also a lot of fun) and simply has to pass through to the charm_tracing lib

Or, perhaps that's better, we keep it outside of the charm concern altogether. I think I have an idea.

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: good point. Will get rid of that.

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: (updated)

lib/charms/tempo_k8s/v0/snapshot.py
```diff
@@ -0,0 +1,321 @@
...
+            contents = cast(
+                Dict[str, str], be._run("secret-get", secret_id, return_output=True, use_json=True)
+            )
+            revision = info["revision"]
+            owner = info["owner"]
+            rotation = info["rotation"]
+            secrets.append(
+                scenario.Secret(
+                    secret_id,
+                    contents={revision: contents},
```
sed-i, line 119: Does this expose the secret? Probably shouldn't?

lib/charms/tempo_k8s/v0/snapshot.py
```diff
@@ -0,0 +1,321 @@
...
+    def get_stored_state(self) -> List[scenario.StoredState]:  # noqa: D102
+        return self._unit_state_db.get_stored_state()
+
+    def get_secrets(self) -> List[scenario.Secret]:  # noqa: D102
+        secrets = []
+        be = self.charm.unit._backend
+        for secret_id in cast(List[str], be._run("secret-ids", return_output=True, use_json=True)):
+            info = cast(
+                Dict[str, Any],
+                be._run("secret-info-get", secret_id, return_output=True, use_json=True),
```
sed-i, line 107: Here and elsewhere there are interactions with the juju controller via hook tools.
It looks like every hook the charm would be pulling everything.
I wonder what performance impact this may have on juju and on charm settling time.

lib/charms/tempo_k8s/v0/snapshot.py
```diff
@@ -0,0 +1,321 @@
...
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+PYDEPS = ["ops-scenario>=6"]
+
+
+class SnapshotBackend:
```
sed-i, line 37: In python 4, classes should have `.__json__` :pray: 
Could pydantic help in the meanwhile?
Otherwise it seems like the raw output of the hook tools would serve you better here.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -58,58 +56,123 @@ def tracer(self) -> opentelemetry.trace.Tracer:
...
-
-        @property
-        def my_tracing_endpoint(self) -> Optional[str]:
-            '''Tempo endpoint for charm tracing'''
-            if self.tracing.is_ready():
-                return self.tracing.otlp_grpc_endpoint()
-            else:
-                return None
-```
+>>>    from charms.tempo_k8s.v0.charm_tracing import trace_charm
```
mmkay, line 60: Minor: `v0` -> `v1` as we're already touching this

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -296,7 +416,7 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
 
         original_event_context = framework._event_context
 
-        _service_name = service_name or self.app.name
+        _service_name = service_name or self.unit.name.replace("/", "-")
```
mmkay, line 419: Not certain if that move is in the right direction. It means that in some scenarios each of the units is seen as a separate service in tempo, right? I'd rather add `unit` as one of the tags of created spans as in case of fast-changing units this could quickly increase the number of services that are seen in tempo search on Grafana to the point of it not being readable.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -337,7 +457,18 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
...
+        attributes = {"juju.dispatch_path": dispatch_path}
+        if SNAPSHOT_CONFIG.active:
+            logger.debug("gathering state...")
+            state = _get_state(self)
+            if state:
+                attributes["state"] = state
+            else:
+                logger.debug(f"could not collect state for {self}")
+
+        logger.debug("initializing root 'charm exec' span")
```
mmkay, line 469: Not certain if that log adds context. It might if we debug-logged the created root span ID - but then we'd need to add the log after starting the span.

scripts/get_charm_states.py
```diff
@@ -0,0 +1,70 @@
...
+app = typer.Typer()
+
+
+def _get_charm_states(target: str, tempo: str, output: Optional[Path]=None):
+    # get trace IDs for this target
+    target_name = target.replace("/", "-")
+    req = requests.get(
+        f"{tempo}/api/search",
+        params={
+            "q": f'{{rootServiceName = "{target_name}"}}'
```
mmkay, line 18: If you agree with not changing the service name to contain the unit (comment in `charm_tracing` above)  you may need to do the split here to search for `rootServiceName` and `unitName`.

mmkay (commented):
Couple more changes on the second round after refactor. I really like moving the entire conversion logic to `snapshot` but @sed-i's comments seem good to check as well. 

IMO changing service name to include unit number might cause some issues on search and generating service maps which we might do later.

lib/charms/tempo_k8s/v0/snapshot.py
PietroPasotti: yes, that's a good point and something we'd need to test. 
Also it's probably a good idea to put that code inside the trace itself, so we can keep an eye on what of an overhead it's having. Will do that.

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: mmm, that's only going to be a problem if:
- we have TONS of units
- we are on a machine model, where unit IDs are not recycled if you respawn them.

I see the danger, however, I think there is no point in doing charm tracing if we don't know exactly which charm trace we are looking at.

If you have many units and you do a blanket search and you get too many results, you should IMHO simply narrow down your query to a range of unit IDs you're interested in

PietroPasotti: or are you suggesting to split service and unit id to two separate labels?

lib/charms/tempo_k8s/v0/snapshot.py
sed-i: Iirc, hook tools already return json, so by calling the _ModelBackend hook tools methods directly you may not need to do much else.

scripts/get_charm_states.py
PietroPasotti: ah, we already have j.uju_unit from the juju topology, we can do that either way. Fixed

lib/charms/tempo_k8s/v1/charm_tracing.py
mmkay: Yes, that was my suggestion - to split service name from the unit ID (which should stay as a label, we want to be able to filter by a specific unit) - unless we want to keep charm traces separate from the workload traces. (btw now that I think about it I wonder if charm traces shouldn't have a label distinguishing them from other traces)

In most cases I don't expect we (or rather the charm developer) are passing unit id as part of the service name to workloads so that they send traces with the right context.  However I might be wrong :)

lib/charms/tempo_k8s/v0/snapshot.py
PietroPasotti: Yes, it looks like most of the time is spent inside `ops.model.Model._run`
image

lib/charms/tempo_k8s/v0/snapshot.py
PietroPasotti: mm, you're saying this because you want to avoid the scenario dependency I think, and I think it's a fair point. And by bypassing the framework too, we can also avoid having to serialize an ops.Relation.
Okay, I'm on board.

PietroPasotti:
this requires some further speccing

PR #81, opened 2024-03-15
Pebble notice tempo ready

PietroPasotti:
This PR implements a pebble notice so that the charm can update its own status once the workload API becomes ready.

### DIFF
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -75,6 +75,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
+    Iterable,
     List,
     Literal,
     MutableMapping,
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -75,7 +75,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
-    Iterable,
     List,
     Literal,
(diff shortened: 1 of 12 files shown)


src/charm.py
```diff
@@ -190,6 +200,12 @@ def _requested_receivers(self) -> Tuple[ReceiverProtocol, ...]:
         requested_receivers = requested_protocols.intersection(set(self.tempo.supported_receivers))
         return tuple(requested_receivers)
 
+    def _on_tempo_pebble_custom_notice(self, event: PebbleNoticeEvent):
+        if event.notice.key == self.tempo.tempo_ready_notice_key:
+            logger.info("pebble api reports ready")
```
mmkay, line 205: We could go down to debug with this one?

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -75,7 +75,6 @@ def __init__(self, *args):
     TYPE_CHECKING,
     Any,
     Dict,
-    Iterable,
     List,
```
mmkay, line 78: Nit: should we bump lib versions for this or just assume it will get released with the next one?

mmkay (approved):
Couple of nitpicks, but overall looks good :)

lib/charms/tempo_k8s/v2/tracing.py
PietroPasotti: oh, I thought I'd rolled that back. Local tox -e lint complained about it.

PR #89, opened 2024-04-16
Tempo client config fix

PietroPasotti:
A number of fixes to tempo from the add-nginx branch
- removed legacy 'tempo' server address from the receiver list
- cleaned up server and receiver concerns in Tempo class
- added a list-receivers action to tempo charm
- added tests to validate receiver publish behaviour on v1 and v2 tracing relation events regardless of ordering


lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -556,6 +564,7 @@ def _get_requested_protocols(relation: Relation):
         try:
             databag = TracingRequirerAppData.load(relation.data[app])
         except (json.JSONDecodeError, pydantic.ValidationError, DataValidationError):
+            logger.exception("What was that error again?")
```
mmkay, line 567: Let's get rid of this leftover. D:

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -585,8 +594,10 @@ def publish_receivers(self, receivers: Sequence[RawReceiver]):
             try:
                 TracingProviderAppData(
                     host=self._host,
+                    external_url=self._external_url,
                     receivers=[
-                        Receiver(port=port, protocol=protocol) for protocol, port in receivers
+                        Receiver(port=port, protocol=protocol, path=f"/{protocol}")
```
mmkay, line 599: Without nginx the path part should be empty for now?

src/charm.py
```diff
@@ -83,9 +83,14 @@ def __init__(self, *args):
         # self._profiling = ProfilingEndpointProvider(
         #     self, jobs=[{"static_configs": [{"targets": ["*:4080"]}]}]
         # )
+        # TODO:
+        #  ingress route provisioning a separate TCP ingress for each receiver if GRPC doesn't work directly
 
-        self._tracing = TracingEndpointProvider(self, host=tempo.host)
-        self._ingress = IngressPerAppRequirer(self, port=self.tempo.tempo_port)
+        self._ingress = IngressPerAppRequirer(self, port=8080, strip_prefix=True)
```
mmkay, line 89: Port needs to be tempo http port again. Without nginx we don't publish anything on 8080.

src/charm.py
```diff
@@ -119,44 +127,48 @@ def legacy_v1_relations(self):
...
         if not self._is_legacy_v1_relation(relation):
-            logger.error(f"relation {relation} is not tracing v1 nor v2. Skipping...")
+            logger.error(f"relation {relation} is not tracing v1. Skipping...")
             return
 
         logger.debug(f"updating legacy v1 relation {relation}")
         # in v1, 'receiver' was called 'ingester'.
         receivers = [
-            tracing_v1.Ingester(protocol=p, port=self.tempo.receiver_ports[p])
+            tracing_v1.Ingester(protocol=p, port=self.tempo.receiver_ports[p], path=f"/{p}")
```
mmkay, line 164: Same here with `path` - I believe in the current idea it's supposed to be empty? I still think that it's good to have it in the protocol as it might become handy in future.

src/charm.py
```diff
@@ -296,6 +319,17 @@ def _on_collect_unit_status(self, e: CollectStatusEvent):
...
 
+    @property
+    def hostname(self) -> str:
+        """Unit's hostname."""
+        return socket.getfqdn()
+
+    def _on_list_receivers_action(self, event: ops.ActionEvent):
+        res = {}
+        for receiver in self._requested_receivers():
+            res[receiver.replace("_", "-")] = f"{self._ingress.url or self.tempo.url}/{receiver}"
```
mmkay, line 330: Not sure how the last part (`/{receiver}`) will act assuming empty paths.

src/tempo.py
```diff
@@ -24,57 +24,56 @@ class Tempo:
...
         # todo allow remapping ports?
+        all_ports = {**self.server_ports}
         return [
             (
-                (f"{service_name_prefix}-{protocol}").replace("_", "-"),
-                self.receiver_ports[protocol],
-                self.receiver_ports[protocol],
+                (f"{service_name_prefix}-{service_name}").replace("_", "-"),
+                all_ports[service_name],
+                all_ports[service_name],
```
mmkay, line 74: Might be a stupid question but now that I think about it: why do we have the same int argument twice here in the tuple?

tests/scenario/test_tracing_legacy.py
```diff
@@ -42,15 +43,15 @@ def test_tracing_v1_endpoint_published(context, base_state):
...
-        '{"protocol": "jaeger_grpc", "port": 14250}]',
-        "host": json.dumps(socket.getfqdn()),
-    }
+    ingesters = json.loads(tracing_out.local_app_data["ingesters"])
+    assert len(ingesters) == NO_RECEIVERS_LEGACY
+
+    for ingester in ingesters:
+        protocol = ingester["protocol"]
+        assert ingester["port"] == Tempo.receiver_ports[protocol]
+        assert ingester["path"] == f"/{protocol}"
```
mmkay, line 52: Here as well I think path might become empty.

tests/scenario/test_tracing_legacy.py
```diff
@@ -60,12 +61,12 @@ def test_tracing_v2_endpoint_published(context, evt_name, base_state):
...
     with charm_tracing_disabled():
         with context.manager(getattr(tracing, f"{evt_name}_event"), state) as mgr:
-            assert len(mgr.charm._requested_receivers()) == 0
+            assert len(mgr.charm._requested_receivers()) == 1
             out = mgr.run()
 
     tracing_out = out.get_relations(tracing.endpoint)[0]
     assert tracing_out.local_app_data == {
-        "receivers": "[]",
+        "receivers": '[{"protocol": "otlp_http", "port": 4318, "path": "/otlp_http"}]',
```
mmkay, line 69: ...and here.

mmkay (commented):
Nice trim! My main concern here is the `path` argument. IMO it might be good to keep it in the protocol definition but I think we shouldn't fill it with protocols and keep it empty as we decided not to add nginx right now.

lib/charms/tempo_k8s/v2/tracing.py
PietroPasotti: hahaha yeah

src/tempo.py
PietroPasotti: IIRC the kubernetesservicepatch thingy (which this is for) takes two ports to allow remapping them.

PR #94, opened 2024-04-25
Add traefik_route to serve tracing endpoints through ingress if it exists

### LINKED ISSUE #82: Make ingestion endpoints available over ingress
### Enhancement Proposal

# Issue

- at the moment, Tempo exposes one port via ingress - its rest APIs port.
- however, each of the ingestion formats Tempo offers is on a different port - that's 8 ports altogether (opentelemetry collector does the same)
- ingress library and traefik in their current form are limited to 1 port per ingress
- without ingestion exposed via ingress, cross-model tracing works only if both models are within the same k8s cluster.
- all charms using charm_tracing use only one protocol - OTLP/HTTP, but we'd eventually like to expose other protocols for workloads as well 

# Ideas

- add a second ingress for OTLP/HTTP to tempo and assume cross-model traces will only be using this protocol for now;
- add ingresses for all the ingestion formats and expose each protocol as a separate ingress (which also means a different URL so potentially we might need to change what we store in the databag);
- modify traefik charm and add another ingress library for multi-port ingress
- use [traefik-route](https://github.com/canonical/traefik-route-k8s-operator)

mmkay:
## Issue
Tempo receiver endpoints were not accessible over traefik as they are bound to different ports. Fixes #82 

## Solution
Use `traefik_route` with changes introduced in [traefik-k8s-operator#325](https://github.com/canonical/traefik-k8s-operator/pull/325) to define ports traefik needs to open.
Also fix creating URLs for tracing consumers that use `external_url` - scheme and port were missing from the created url.

## Context
This change is needed for cross-model traces to work. They cannot access internal endpoints of a different k8s cluster . No traces were also coming from machine models.

## Testing Instructions
Run the following bundle and verify endpoints (example, port 3200, `/metrics`, port 4318 (expected prometheus metrics), `/v1/traces` (expected HTTP code 405)) are accessible on traefik's external IP.
```
bundle: kubernetes
applications:
  tempo:
    charm: local:tempo-k8s-1
    scale: 1
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
  traefik-k8s:
    charm: traefik-k8s
    channel: latest/edge
    revision: 184
    base: ubuntu@20.04/stable
    resources:
      traefik-image: 159
    scale: 1
    constraints: arch=amd64
    storage:
      configurations: kubernetes,1,1024M
relations:
- - tempo:tracing
  - traefik-k8s:tracing
- - traefik-k8s:traefik-route
  - tempo:ingress
```

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -58,6 +58,9 @@ requires:
   ingress:
     interface: traefik_route
     limit: 1
+    description: |
+      Adds an ingressed endpoint used for cross-model communication.
+      Uses `traefik_route` to open ports on Traefik host for tracing ingesters.
 
 storage:
   data:
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -59,7 +59,8 @@ requires:
     interface: traefik_route
     limit: 1
     description: |
-      Adds an ingressed endpoint used for cross-model communication.
+      Ingress integration for Tempo server and Tempo receiver endpoints,
+      so that cross-model workloads can send their traces to Tempo through the ingress.
       Uses `traefik_route` to open ports on Traefik host for tracing ingesters.
 
 storage:
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -56,7 +56,7 @@ requires:
   logging:
     interface: loki_push_api
   ingress:
-    interface: ingress
+    interface: traefik_route
     limit: 1
 
 storage:
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -104,7 +104,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
 
 PYDEPS = ["pydantic"]
 
@@ -822,7 +822,7 @@ class TracingEndpointRequirer(Object):
         receiver = receivers[0]
         # if there's an external_url argument (v2.5+), use that. Otherwise, we use the tempo local fqdn
         if app_data.external_url:
-            url = app_data.external_url
+            url = f"http://{app_data.external_url}"
         else:
             # FIXME: if we don't get an external url but only a
             #  hostname, we don't know what scheme we need to be using. ASSUME HTTP
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -590,7 +590,7 @@ class TracingEndpointProvider(Object):
             try:
                 TracingProviderAppData(
                     host=self._host,
-                    external_url=self._external_url,
+                    external_url=f"http://{self._external_url}" if self._external_url else None,
                     receivers=[
                         Receiver(port=port, protocol=protocol) for protocol, port in receivers
                     ],
@@ -822,7 +822,7 @@ class TracingEndpointRequirer(Object):
         receiver = receivers[0]
         # if there's an external_url argument (v2.5+), use that. Otherwise, we use the tempo local fqdn
         if app_data.external_url:
-            url = f"http://{app_data.external_url}"
+            url = f"{app_data.external_url}:{receiver.port}"
         else:
             # FIXME: if we don't get an external url but only a
             #  hostname, we don't know what scheme we need to be using. ASSUME HTTP
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -308,6 +308,9 @@ class TracingProviderAppData(DatabagModel):  # noqa: D101
     external_url: Optional[str] = None
     """Server url. If an ingress is present, it will be the ingress address."""
 
+    internal_scheme: Optional[str] = None
+    """Scheme for internal communication. If it is present, it will be protocol accepted by the provider."""
+
 
 class TracingRequirerAppData(DatabagModel):  # noqa: D101
     """Application databag model for the tracing requirer."""
@@ -495,6 +498,7 @@ class TracingEndpointProvider(Object):
         host: str,
         external_url: Optional[str] = None,
         relation_name: str = DEFAULT_RELATION_NAME,
+        internal_scheme: Optional[str] = "http",
     ):
         """Initialize.
 
@@ -525,6 +529,7 @@ class TracingEndpointProvider(Object):
         self._host = host
         self._external_url = external_url
         self._relation_name = relation_name
+        self._internal_scheme = internal_scheme
         self.framework.observe(
             self._charm.on[relation_name].relation_joined, self._on_relation_event
         )
@@ -594,6 +599,7 @@ class TracingEndpointProvider(Object):
                     receivers=[
                         Receiver(port=port, protocol=protocol) for protocol, port in receivers
                     ],
+                    internal_scheme=self._internal_scheme,
                 ).dump(relation.data[self._charm.app])
 
             except ModelError as e:
@@ -824,9 +830,11 @@ class TracingEndpointRequirer(Object):
         if app_data.external_url:
             url = f"{app_data.external_url}:{receiver.port}"
         else:
-            # FIXME: if we don't get an external url but only a
-            #  hostname, we don't know what scheme we need to be using. ASSUME HTTP
-            url = f"http://{app_data.host}:{receiver.port}"
+            if app_data.internal_scheme:
+                url = f"{app_data.internal_scheme}://{app_data.host}:{receiver.port}"
+            else:
+                # if we didn't receive a scheme (old provider), we assume HTTP is used
+                url = f"http://{app_data.host}:{receiver.port}"
 
         if receiver.protocol.endswith("grpc"):
             # TCP protocols don't want an http/https scheme prefix
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -498,7 +498,7 @@ class TracingEndpointProvider(Object):
         host: str,
         external_url: Optional[str] = None,
         relation_name: str = DEFAULT_RELATION_NAME,
-        internal_scheme: Optional[str] = "http",
+        internal_scheme: Optional[Literal["http", "https"]] = "http",
     ):
         """Initialize.
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -33,6 +33,7 @@ from ops.charm import (
 )
 from ops.main import main
 from ops.model import ActiveStatus, MaintenanceStatus, Relation, WaitingStatus
+
 from tempo import Tempo
 
 logger = logging.getLogger(__name__)
@@ -86,7 +87,11 @@ class TempoCharm(CharmBase):
 
         self._ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
         self._tracing = TracingEndpointProvider(
-            self, host=self.tempo.host, external_url=self._ingress.external_host
+            # TODO set internal_scheme based on whether TLS is enabled
+            self,
+            host=self.tempo.host,
+            external_url=self._ingress.external_host,
+            internal_scheme="http",
         )
 
         self.framework.observe(
--- a/src/charm.py
+++ b/src/charm.py
@@ -393,7 +393,7 @@ class TempoCharm(CharmBase):
                     "rule": "ClientIP(`0.0.0.0/0`)",
                 }
                 http_services[
-                    f"juju-{self.model.name}-{self.model.app.name}-{sanitized_protocol}"
+                    f"juju-{self.model.name}-{self.model.app.name}-service-{sanitized_protocol}"
                 ] = {"loadBalancer": {"servers": [{"url": f"http://{self.hostname}:{port}"}]}}
         return {
             "tcp": {
--- a/src/charm.py
+++ b/src/charm.py
@@ -22,12 +22,14 @@ from charms.tempo_k8s.v2.tracing import (
     RequestEvent,
     TracingEndpointProvider,
 )
-from charms.traefik_k8s.v2.ingress import IngressPerAppRequirer
+from charms.traefik_route_k8s.v0.traefik_route import TraefikRouteRequirer
 from ops.charm import (
     CharmBase,
     CollectStatusEvent,
+    HookEvent,
     PebbleNoticeEvent,
     RelationEvent,
+    RelationJoinedEvent,
     WorkloadEvent,
 )
 from ops.main import main
@@ -86,13 +88,17 @@ class TempoCharm(CharmBase):
         # TODO:
         #  ingress route provisioning a separate TCP ingress for each receiver if GRPC doesn't work directly
 
-        self._ingress = IngressPerAppRequirer(
-            self, port=self.tempo.tempo_server_port, strip_prefix=True
+        self._ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
+        self._tracing = TracingEndpointProvider(
+            self, host=self.tempo.host, external_url=self._ingress.external_host
         )
 
-        self._tracing = TracingEndpointProvider(
-            self, host=self.tempo.host, external_url=self._ingress.url
-        )
+        self.framework.observe(self.on["ingress"].relation_joined, self._configure_ingress)
+        self.framework.observe(self.on.leader_elected, self._configure_ingress)
+        self.framework.observe(self.on.config_changed, self._configure_ingress)
+        self.framework.observe(self._ingress.on.ready, self._on_ingress_ready)  # pyright: ignore
+        # TODO there's no revoked ingress action with traefik_route?
+        #self.framework.observe(self._ingress.on.revoked, self._on_ingress_revoked)
 
         self.framework.observe(self.on.tempo_pebble_ready, self._on_tempo_pebble_ready)
         self.framework.observe(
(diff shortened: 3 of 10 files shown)


src/charm.py
```diff
@@ -86,14 +87,18 @@ def __init__(self, *args):
...
-        )
-
+        self._ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
         self._tracing = TracingEndpointProvider(
-            self, host=self.tempo.host, external_url=self._ingress.url
+            self, host=self.tempo.host, external_url=self._ingress.external_host
         )
 
+        self.framework.observe(self.on["ingress"].relation_joined, self._configure_ingress)
+        self.framework.observe(self.on.leader_elected, self._configure_ingress)
```
PietroPasotti, line 96: all observe calls should be: 

`self.on.event, self._on_event`

and in `def self._on_event` you can do whatever you like. This is easier to maintain and follow.

src/charm.py
```diff
@@ -86,14 +87,18 @@ def __init__(self, *args):
...
+        self.framework.observe(self.on.config_changed, self._configure_ingress)
+        self.framework.observe(self._ingress.on.ready, self._on_ingress_ready)  # pyright: ignore
+        # TODO there's no revoked ingress action with traefik_route?
```
PietroPasotti, line 99: no, you have to listen to broken/departed separately. Or contribute one to traefik-route I guess

src/charm.py
```diff
@@ -86,14 +87,18 @@ def __init__(self, *args):
...
+        # self.framework.observe(self._ingress.on.revoked, self._on_ingress_revoked)
```
PietroPasotti, line 100:
```suggestion
       self.framework.observe(self._ingress.on.broken, self._on_ingress_broken)
```

src/charm.py
```diff
@@ -86,14 +87,18 @@ def __init__(self, *args):
...
-
+        self._ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
         self._tracing = TracingEndpointProvider(
-            self, host=self.tempo.host, external_url=self._ingress.url
+            self, host=self.tempo.host, external_url=self._ingress.external_host
         )
 
+        self.framework.observe(self.on["ingress"].relation_joined, self._configure_ingress)
+        self.framework.observe(self.on.leader_elected, self._configure_ingress)
+        self.framework.observe(self.on.config_changed, self._configure_ingress)
```
PietroPasotti, line 97: why do we need this? Tempo has no config at the moment.

src/charm.py
PietroPasotti: secondly, why do we need to observe the leader-elected hook? 
I guess it's because traefik-route goes through app data, and if your leader dies and you are elected, you need to take lead of traefik-route. Can you add a comment?

src/charm.py
```diff
@@ -121,6 +124,34 @@ def _is_legacy_v1_relation(self, relation):
 
         return True
 
+    def _configure_ingress(self, event: HookEvent) -> None:
+        """Set up ingress if a relation is joined, config changed, or a new leader election.
```
PietroPasotti, line 128:
```suggestion
        """Make sure the traefik route relation data is up to date.
```

src/charm.py
```diff
@@ -121,6 +124,34 @@ def _is_legacy_v1_relation(self, relation):
...
+
+        Args:
+            event: a :class:`HookEvent` to signal a change we may need to respond to.
+        """
+        if not self.unit.is_leader():
+            return
+
+        # If it's a RelationJoinedEvent, set it in the ingress object
+        if isinstance(event, RelationJoinedEvent):
+            self._ingress._relation = event.relation
```
PietroPasotti, line 143: it's very weird that it wasn't there to begin with. Where have you being seeing this? 
If it's Harness, this is an artifact due to the charm not being reinitialized on each event and should be fixed in test code, not charm code.

src/charm.py
```diff
@@ -121,6 +124,34 @@ def _is_legacy_v1_relation(self, relation):
...
+
+        # No matter what, check readiness -- this blindly checks whether `ingress._relation` is not
+        # None, so it overlaps a little with the above, but works as expected on leader elections
+        # and config-change
+        if self._ingress.is_ready():
+            self._update_tracing_v1_relations()
```
PietroPasotti, line 149: it doesn't make sense to update the tracing relations unless we've `submit_to_traefik` already in a previous event. 
First we submit to traefik, then traefik replies with 'here's your url', then we can update the tracing relations. 
So on ingress ready you want to update tracing v1 and v2 relations, but on ingress CREATED or joined, that's where you want to submit_to_traefik

Or maybe I'm misremembering the semantics of traefik_route.on.ready.

Please add a test for this :)

charmcraft.yaml
```diff
@@ -45,7 +45,7 @@ requires:
   logging:
     interface: loki_push_api
   ingress:
-    interface: ingress
+    interface: traefik_route
```
PietroPasotti, line 48: add a description too please

PietroPasotti (changes requested):
Looks good! However I'd really like to see some tests.
At least a couple of unittests for databag contents and endpoints based on whether there is an ingress relation.

Secondly, wondering if we should be renaming the 'ingress' endpoint to `traefik-route` to follow the convention of having endpoints and interfaces similarly named. 
Also, if some day we choose to provide this kind of multi-route, hybrid ingress via traefik and upgrade the ingress relation accordingly, we can swap them out without the `ingress` name being taken by something that isn't quite ingress.

src/charm.py
mmkay: This is a bit confusing as I tried that and it looks like there's only `ready` for `TraefikRouteRequirer`? https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_route_k8s/v0/traefik_route.py#L119

src/charm.py
mmkay: Changed it according to the guidelines and added the requested comment.

src/charm.py
mmkay: True, removed.

src/charm.py
mmkay: That is based on a similar issue in grafana: https://github.com/canonical/grafana-k8s-operator/blob/main/src/charm.py#L353

charmcraft.yaml
mmkay: Added.

src/charm.py
```diff
@@ -104,8 +110,6 @@ def __init__(self, *args):
         self.framework.observe(self.on.tracing_relation_joined, self._on_tracing_relation_joined)
         self.framework.observe(self.on.tracing_relation_changed, self._on_tracing_relation_changed)
         self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
-        self.framework.observe(self._ingress.on.ready, self._on_ingress_ready)
-        self.framework.observe(self._ingress.on.revoked, self._on_ingress_revoked)
```
sed-i, line 108: We don't need revoked anymore?
E.g. to update the server url we advertize over tracing relations?

src/charm.py
```diff
@@ -121,6 +125,25 @@ def _is_legacy_v1_relation(self, relation):
 
         return True
 
+    def _configure_ingress(self, event: HookEvent) -> None:
+        """Make sure the traefik route relation data is up to date."""
+        if not self.unit.is_leader():
+            return
```
sed-i, line 131: Do you mean this is "ingress to leader"?
Is there any cleanup we would need to do on `leader-settings-changed`?

src/charm.py
```diff
@@ -315,9 +348,66 @@ def hostname(self) -> str:
...
+    def _static_ingress_config(self) -> dict:
+        entry_points = {}
+        for protocol, port in self.tempo.all_ports.items():
+            sanitized_protocol = protocol.replace("_", "-")
+            entry_points[sanitized_protocol] = {"address": f":{port}"}
+
+        return {"entryPoints": entry_points}
+
+    @property
+    def _ingress_config(self) -> dict:
```
sed-i, line 366: Would it make sense to extract this logic into an ops-independent module?

src/charm.py
mmkay: Yup, seems like it was a testing artifact. Removed it now.

src/charm.py
mmkay: You're correct - changed it to match what you said.

src/charm.py
mmkay: We do have `_on_leader_elected` for updating the relations: https://github.com/canonical/tempo-k8s-operator/blob/4f97ae8185d46062b48cbf9e5e101dce7deffff1/src/charm.py#L178

src/charm.py
mmkay: It seems that we don't have a `revoked` event to observe in case of `TraefikRouteRequirer`: https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_route_k8s/v0/traefik_route.py#L119

src/charm.py
mmkay: There's a similar section in grafana-k8s-operator: https://github.com/canonical/grafana-k8s-operator/blob/main/src/charm.py#L1325

I think that yes, it might make sense, but as a separate issue.

mmkay:
@PietroPasotti: regarding this part:

> Secondly, wondering if we should be renaming the 'ingress' endpoint to traefik-route to follow the convention of having endpoints and interfaces similarly named.
Also, if some day we choose to provide this kind of multi-route, hybrid ingress via traefik and upgrade the ingress relation accordingly, we can swap them out without the ingress name being taken by something that isn't quite ingress.

I don't mind changing the interface name but I was basing the current approach on that Grafana also uses `ingress` in the same manner: 

https://github.com/canonical/grafana-k8s-operator/blob/main/charmcraft.yaml#L54

As it was the only o11y charm that uses `traefik_route` that I know of, I chose the same pattern as it is fairly transparent to the charm users - they want to relate to Traefik and they're (mostly) used that the protocol they would use is named "ingress". I can however change that if you think having `ingress` free for future use is useful :)

src/charm.py
mmkay: Created an issue in traefik: https://github.com/canonical/traefik-k8s-operator/issues/345

IMO Grafana doesn't handle it either at the moment so this can be dealt with separately.

src/charm.py
mmkay: Created an issue in traefik: https://github.com/canonical/traefik-k8s-operator/issues/345

IMO Grafana doesn't handle it either at the moment so this can be dealt with separately.

tests/scenario/test_ingressed_tracing.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_external_url_present(context, base_state):
+    # WHEN ingress is related with external_host
+    tracing = Relation("tracing", remote_app_data={"receivers": "[]"})
+    ingress = Relation("ingress", remote_app_data={"external_host": "1.2.3.4", "scheme": "http"})
+    state = base_state.replace(relations=[tracing, ingress])
+
+    with charm_tracing_disabled():
+        with context.manager(getattr(tracing, "created_event"), state) as mgr:
```
PietroPasotti, line 21: if you're not planning to use mgr.charm, you should:

```suggestion
        context.run(getattr(tracing, "created_event"), state)
        ```

File: tests/scenario/test_tracing_requirer.py [hunk 13]
```diff
@@ -61,6 +61,42 @@ def test_requirer_api(context):
...
+    state = State(leader=True, relations=[tracing])
+
+    # THEN get_endpoint uses external URL instead of the host
+    with charm_tracing_disabled():
+        with context.manager(tracing.changed_event, state) as mgr:
+            charm = mgr.charm
+            assert (
+                charm.tracing.get_endpoint("otlp_grpc") == f"{external_url.split('://')[1]}:4317"
+            )
+            assert charm.tracing.get_endpoint("otlp_http") == f"{external_url}:4318"
```
PietroPasotti, line 87: it's a bit brittle to use hardcoded port ids in tests. 
Use `Tempo._receiver_ports...`

```suggestion
            for proto in ["otlp_http", ...]:
                assert charm.tracing.get_endpoint(proto) == f"{external_url}:{Tempo._receiver_ports[proto]"
```

src/charm.py
```diff
@@ -146,6 +162,23 @@ def _on_tracing_relation_changed(self, e: RelationEvent):
...
             self._publish_v1_data(e.relation)
 
+    def _on_ingress_relation_created(self, e: RelationEvent):
+        if self._ingress.is_ready():
+            self._ingress.submit_to_traefik(
+                self._ingress_config, static=self._static_ingress_config
+            )
+
+    def _on_ingress_relation_joined(self, e: RelationEvent):
+        self._ingress._relation = e.relation
```
PietroPasotti, line 172: what's going on here?

src/charm.py
```diff
@@ -146,6 +162,23 @@ def _on_tracing_relation_changed(self, e: RelationEvent):
         if not self._tracing.is_v2(e.relation):
             self._publish_v1_data(e.relation)
 
+    def _on_ingress_relation_created(self, e: RelationEvent):
+        if self._ingress.is_ready():
```
PietroPasotti, line 166: duplicated code between created and joined handler, consolidate?

PietroPasotti: IIUC you could just call `_configure_ingress()` and be done with it, no?

src/charm.py
```diff
@@ -121,6 +125,18 @@ def _is_legacy_v1_relation(self, relation):
...
+    def _configure_ingress(self, _) -> None:
+        """Make sure the traefik route and tracing relation data are up to date."""
+        if not self.unit.is_leader():
+            return
+
+        if self._ingress.is_ready():
+            self._ingress.submit_to_traefik(
+                self._ingress_config, static=self._static_ingress_config
+            )
+            self._update_tracing_v1_relations()
```
PietroPasotti, line 137: you could do this conditionally on there being an ingress url already. Pointless to update relations if traefik hasn't sent you an url back yet. We should only do this on ingress ready in fact
`if self._ingress.url: ...`

PietroPasotti (changes requested):
looking good! a few things that we need to check :)

src/charm.py
mmkay: True - changed that.

src/charm.py
mmkay: You were right in another conversation - this looks like a Harness issue inspired by grafana-k8s-operator workaround. Rewrote the unit test to be a scenario test for the same test case and removed the line as it passes.

charmcraft.yaml
```diff
@@ -56,8 +56,11 @@ requires:
...
     limit: 1
+    description: |
+      Adds an ingressed endpoint used for cross-model communication.
```
PietroPasotti, line 62:
```suggestion
      Ingress integration for Tempo server and Tempo receiver endpoints, so that cross-model workloads can send their traces to Tempo through the ingress.
```

src/charm.py
```diff
@@ -83,16 +83,20 @@ def __init__(self, *args):
...
         #     self, jobs=[{"static_configs": [{"targets": ["*:4080"]}]}]
         # )
-        # TODO:
-        #  ingress route provisioning a separate TCP ingress for each receiver if GRPC doesn't work directly
 
-        self._ingress = IngressPerAppRequirer(
-            self, port=self.tempo.tempo_server_port, strip_prefix=True
+        self._ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
+        self._tracing = TracingEndpointProvider(
+            self, host=self.tempo.host, external_url=self._ingress.external_host
```
PietroPasotti, line 89: we're introducing an inconsistency here: 
- external url will be http or https depending on whether traefik has TLS active
- host will be a raw hostname, and the tracing lib will by default add http in front of it.

How about: 
- we rename host to internal_url (with a backwards-compat deprecation warning to give us time to adjust all dependent charms)
- we generate the internal_url from the tempo charm, not the charm lib, so the charm remains responsible of determining whether we should be using https or http. 

Other alternative would be to add to the lib an `internal_scheme` arg that the charm lib could use instead of the hardcoded `http`.

PietroPasotti (changes requested):
Mainly the provider endpoint api needs a final touch IMHO

src/charm.py
mmkay: True - I liked your first idea (host rename) a bit better, but I feel it's causing a bit more backwards compatibility issues than adding `internal_scheme` to the databag, so I went with the second option. There's still fallback to `http://` in case newer charm uses an older Tempo to send traces to as well.

charmcraft.yaml
mmkay: Done, thanks!

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -495,6 +498,7 @@ def __init__(
         host: str,
         external_url: Optional[str] = None,
         relation_name: str = DEFAULT_RELATION_NAME,
+        internal_scheme: Optional[str] = "http",
```
PietroPasotti, line 501:
```suggestion
        internal_scheme: Optional[Literal["http", "https"]] = "http",
```

PietroPasotti (approved):
one final typing suggestion, then we can go!
Only concern I'm left with is the situation where Traefik has TLS, tempo doesn't: external url will be https, internal one will be http.

Will ingestion work via both?

Once you're convinced this is fine, we're good :)

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -590,10 +595,11 @@ def publish_receivers(self, receivers: Sequence[RawReceiver]):
             try:
                 TracingProviderAppData(
                     host=self._host,
-                    external_url=self._external_url,
+                    external_url=f"http://{self._external_url}" if self._external_url else None,
```
PietroPasotti, line 598: ok, so this is definitely wrong.
What's happening is: traefik-route is giving us a 'scheme' but we're completely ignoring it.

image

what we should be doing is:
image

and then:

image

PietroPasotti (changes requested):
there's something wrong with http/https and ingress after all

lib/charms/tempo_k8s/v2/tracing.py
PietroPasotti: see #98 for my current progress on this

PietroPasotti (approved):
what an emotional rollercoaster

PR #97, opened 2024-04-26
Set total block retention to 30 days

mmkay:
## Issue
Traces seem to disappear after a longer period.

## Solution
Set default `block_retention` to 30 days. Set also `compacted_block_retention` that says how long blocks should stay around after they've been compacted

## Context
See https://github.com/grafana/tempo/blob/main/modules/compactor/config.go#L38 and https://grafana.com/docs/tempo/latest/configuration/#compactor for context.

## Testing Instructions
- deploy tempo with something that posts traces
- keep it running for more than 1h
- you should still see the original traces

### DIFF
--- a/src/tempo.py
+++ b/src/tempo.py
@@ -191,8 +191,9 @@ class Tempo:
                     "compaction_window": "1h",
                     # maximum size of compacted blocks
                     "max_compaction_objects": 1000000,
-                    "block_retention": "1h",
-                    "compacted_block_retention": "10m",
+                    # total trace retention
+                    "block_retention": "30d",
+                    "compacted_block_retention": "1h",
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -60,7 +60,7 @@ commands =
 [testenv:scenario]
 description = Run scenario tests
 deps =
-    pytest
+    pytest<8.2.0 # https://github.com/pytest-dev/pytest/issues/12263
```
PietroPasotti, line 63: I think we only need this in integration, it's a matter of incompatibility with pytest-operator. The other envs should be unaffected.

PietroPasotti (approved):
merge away

PR #98, opened 2024-04-29
TLS integration for tempo endpoints and server

PietroPasotti:
TLS integration support for Tempo charm

In principle ready for review, but it depends on ingress #94 

Testing instructions:

```
juju add-model test-tempo-tls
juju deploy cos-lite --trust
charmcraft pack
jhack deploy
jhack imatrix fill
```

you should be able to see charm traces in grafana for all the charms.
verify that all receiver endpoints are using https:

```
juju run tempo/0 list-receivers
```


michaeldmitry:
if we enable tls for tempo and another charm for charm tracing
and then we disable tls on tempo and the other charm, App data will not change (internal_scheme will still remain as https) and that will fail non-tls communication
is this a valid scenario to cover?

michaeldmitry:
@PietroPasotti  @mmkay 
when certificates are pushed into the workload container, its an ephemeral volume
so, in case of pod crash and restart, the certificates will be no longer there. Once tempo is back up, it will run as non-tls, but all databags are expected to send to a tls endpoint, so tls is now broken.

prometheus handles that https://github.com/canonical/prometheus-k8s-operator/blob/main/src/charm.py#L546

mmkay (approved):
What an odyssey.

PR #99, opened 2024-04-29
use pytest version < 8.2.0

michaeldmitry:
## Issue
When using with the most recent version of pytest 8.2.0, asyncio-0.21.1 is used which breaks the fixtures with error: AttributeError: 'FixtureDef' object has no attribute 'unittest'.
https://github.com/charmed-kubernetes/pytest-operator/issues/131

## Solution
Use pytest version < 8.2.0 for our tests until upstream is fixed.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -48,7 +48,7 @@ commands =
 [testenv:unit]
 description = Run unit tests
 deps =
-    pytest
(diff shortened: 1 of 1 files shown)


dstathis (approved):
+1 For linking to the issue.

PR #106, opened 2024-05-08
Verify and double-check loki alert rules

michaeldmitry:
## Issue
Missing Loki alert rules.

## Solution
Added an alert rule to notify if error logs are triggered in tempo.

## Context
1. promtail will not push to loki, as port 9095 is already used by Tempo. Will be solved in https://github.com/canonical/tempo-k8s-operator/pull/98
2. loki is not able to push alerts to alert manager. Will be solved in https://github.com/canonical/loki-k8s-operator/issues/396


src/loki_alert_rules/error_level_logs_found.rule
```diff
@@ -0,0 +1,7 @@
+alert: ErrorLevelLogsFound
+expr: rate({%%juju_topology%%} |= "level=error"[1m]) > 0
+for: 45s
```
mmkay, line 3: 45 seconds seems like quite a short window especially that we look at logs over the last minute. How about a longer time period, like 10m like in the example here? https://grafana.com/docs/loki/latest/alert/#example

src/loki_alert_rules/error_level_logs_found.rule
```diff
@@ -0,0 +1,7 @@
+alert: ErrorLevelLogsFound
+expr: rate({%%juju_topology%%} |= "level=error"[1m]) > 0
```
mmkay, line 2: Wondering whether any error should be enough to trigger an alert or maybe if we should have a percentage-based check. So something like "if 5% of logs are errors, create an alert". What do you think?

michaeldmitry: Following the same approach we did for Prometheus, having it as a percentage to leave a margin for error seems more reasonable

mmkay (approved):

PR #108, opened 2024-05-23
Clean up tracing databag in v2

### LINKED ISSUE #105: Clean up tracing databag in v3
### Enhancement Proposal

In tracing v2, we made several choices about the structure of the databag. Provider side exposes:

```
- host - used to build internal URL. Doesn't contain the scheme
- receivers - list of supported receivers with their ports
- external_url - when ingress is there, it's the ingressed URL, it contains the scheme
- internal_scheme - optional, used in combination with host (and receivers) to build the path. Can be either 'http' or 'https'
```

This structure has its quirks as we wanted to keep as backwards compatible as possible:
- there's a difference between building tracing URL for ingressed and non-ingressed scenario, causing unnecessary complexity in `tracing.get_endpoint` logic
- we have fields that are conditional

We could do one of the following:

Option 1: 
- the databag contains only receivers. Each receiver contains a full URL, with scheme and port (however without the exact API path - some instrumentation libraries add it themselves). If Tempo is behind ingress, it is the ingressed URL, otherwise the internal one. This means that in all relations (regardless if they're in-model or not), if ingress is there, it is the only advertised option to pass traces.

Option 2:
- the databag contains receivers like it does now, but there are `external_url` and `internal_url` fields, both containing scheme. `get_endpoint` then appends port number and URL that belongs to the right protocol.

michaeldmitry:
## Issue
Fixes #105

## Solution
- update provider app databag to follow https://github.com/canonical/charm-relation-interfaces/pull/150
- update minor version for tracing library (2.6)
- removed tracing library v1 support
- fix UTs, Scenario tests, Integration tests

## Testing instructions

with tempo charm repo root as CWD:
```
juju add-model tempo-clite-test
juju deploy cos-lite --channel edge --trust

git fetch
git checkout OPENG-2399
jhack sync "*" --touch ./lib/charms/tempo_k8s/v2/tracing.py

charmcraft pack
jhack deploy 

jhack imatrix fill
```

now go to the grafana dashboard and ensure you can see charm traces for all apps (and grafana workload)

add tls:
```
juju deploy self-signed-certificates ssc
jhack imatrix fill
```

verify you can still see the new charm traces for all apps (and grafana workload)

Issues:
traefik gets stuck because we removed the deprecated api that traefik was still relying on. see https://github.com/canonical/traefik-k8s-operator/pull/364

Same for grafana: https://github.com/canonical/grafana-k8s-operator/pull/333

Catalogue has broken tls charm tracing
https://github.com/canonical/catalogue-k8s-operator/pull/102


src/tempo.py
```diff
@@ -62,6 +62,7 @@ class Tempo:
     def __init__(
         self,
         container: ops.Container,
+        charm: Optional[object] = None,
```
mmkay, line 65: Would it be possible to pass just the ingress object? That could make it less dependent on having an object named exactly `ingress` in the charm.

mmkay (commented):
I like how much complexity we're removing with these changes. There's a few smaller suggestions that I have but overall I think this is heading in the right direction.

tox.ini
```diff
@@ -105,5 +105,5 @@ deps =
 commands =
     charm: pyright --pythonversion 3.8 {[vars]src_path}
     lib: pyright --pythonversion 3.8 {[vars]lib_path}
-    lib: /usr/bin/env sh -c 'for m in $(git diff main --name-only {[vars]lib_path}); do if ! git diff main $m | grep -q "+LIBPATCH\|+LIBAPI"; then echo "You forgot to bump the version on $m!"; exit 1; fi; done'
+    lib: /usr/bin/env sh -c 'for m in $(git diff main --name-only {[vars]lib_path}); do if git ls-tree -r HEAD --name-only | grep -q "^$m$"; then  if ! git diff main $m | grep -q "+LIBPATCH\|+LIBAPI"; then echo "You forgot to bump the version on $m!"; exit 1; fi; fi; done'
```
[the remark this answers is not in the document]
michaeldmitry: this check fails since it sees a change in the library file v1, although the change is that it got deleted. So, I added a check to not error out if the file has been deleted in the branch.

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -595,12 +614,15 @@ def publish_receivers(self, receivers: Sequence[RawReceiver]):
         for relation in self.relations:
```
michaeldmitry, line 614: @PietroPasotti @mmkay that would would update all relations app data with "recievers" they didn't necessarily subscribe to. Is this an intended behavior?  I don't think that is correct

mmkay: If I recall right I had the same discussion with @PietroPasotti before and I think the result was that Tempo should expose any receiver they have. Correct me if I'm wrong @PietroPasotti  :)

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -113,6 +114,16 @@ def __init__(self, *args):
 DEFAULT_RELATION_NAME = "tracing"
 RELATION_INTERFACE_NAME = "tracing"
 
+ReceiverProtocolType = {
```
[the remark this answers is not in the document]
mmkay: Discussion result: we don't need to remove it as it's fairly unlikely to change. This comment can be then disregarded.

lib/charms/tempo_k8s/v2/tracing.py
michaeldmitry: Observation:
When a tracing relation is removed, the endpoints/protocols it subscribed to, still exists in the databag of *other* charm relations

PietroPasotti: that's weird. We're probably missing a relation-removed observer

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -121,13 +132,9 @@ def __init__(self, *args):
...
     "otlp_grpc",
     "otlp_http",
-    # "jaeger_grpc",
-    "jaeger_thrift_compact",
-    "jaeger_thrift_http",
-    "jaeger_thrift_binary",
 ]
 
-RawReceiver = Tuple[ReceiverProtocol, int]
+RawReceiver = Tuple[ReceiverProtocol, str]
```
PietroPasotti, line 137:
```suggestion
RawReceiver = Tuple[ReceiverProtocol, str]
"""Helper type. A raw receiver is defined as a tuple consisting of the protocol name, and the (external, if available), (secured, if available) resolvable server url."""
```

lib/charms/tempo_k8s/v2/tracing.py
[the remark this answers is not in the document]
michaeldmitry: ReceiverProtocolType is supposed to be exposed outside and used by tempo.py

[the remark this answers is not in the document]
PietroPasotti: oh fair enough, then let's keep it public. Still snake_case instead of CamelCase, and rename it, as it's not a type.

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -121,16 +122,34 @@ def __init__(self, *args):
...
 
 
+class TransportProtocolType(str, enum.Enum):
+    """Receiver Type."""
+
+    http = "http"
+    grpc = "grpc"
+
+
+ReceiverProtocolType = {
```
PietroPasotti, line 142:
```suggestion
receiver_protocol_to_transport_protocol = {
```

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -289,27 +308,79 @@ def dump(self, databag: Optional[MutableMapping] = None, clear: bool = True):
...
+        )
+
+
+class Receiver(BaseModel):
+    """Specification of an active receiver."""
+
+    protocol: ProtocolType = Field(..., description="Receiver protocol name and type.")
+    url: str = Field(
+        ...,
+        description="URL at which the receiver is reachable. If there's an ingress, it would be the external URL. Otherwise, it would be the service's fqdn or internal IP",
```
PietroPasotti, line 366:
```suggestion
        description="URL at which the receiver is reachable. If there's an ingress, it would be the external URL. Otherwise, it would be the service's fqdn or internal IP. If the protocol type is grpc, the url will not contain a scheme.",
```

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -289,27 +308,79 @@ def dump(self, databag: Optional[MutableMapping] = None, clear: bool = True):
...
+        ],
+    )
+
+
+class TracingProviderAppData(DatabagModel):  # noqa: D101
+    """Application databag model for the tracing provider."""
+
+    receivers: List[Receiver] = Field(
+        ...,
+        description="A list of enabled receivers in the form of the protocol they use and their resolvable server url.",
```
PietroPasotti, line 382: The form is already defined in the models, no need to reword it here

```suggestion
        description="List of all receivers enabled on the tracing provider.",
```

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -595,12 +656,17 @@ def publish_receivers(self, receivers: Sequence[RawReceiver]):
...
-                    host=self._host,
-                    external_url=self._external_url or None,
                     receivers=[
-                        Receiver(port=port, protocol=protocol) for protocol, port in receivers
+                        Receiver(
+                            url=url,
+                            protocol=ProtocolType(
+                                name=protocol,
+                                type=ReceiverProtocolType.get(protocol)
+                                or TransportProtocolType(""),
```
PietroPasotti, line 665: what's the `TransportProtocolType("")` about? 
I'd rather do a hard `type=ReceiverProtocolType[protocol]` and let a keyerror raise.
If someone is passing an unknown protocol, we should know that sooner rather than later.

src/charm.py
```diff
@@ -125,13 +112,11 @@ def __init__(self, *args):
...
         self.framework.observe(self.on.update_status, self._on_update_status)
         self.framework.observe(self.tracing.on.request, self._on_tracing_request)
-        self.framework.observe(self.on.tracing_relation_created, self._on_tracing_relation_created)
-        self.framework.observe(self.on.tracing_relation_joined, self._on_tracing_relation_joined)
-        self.framework.observe(self.on.tracing_relation_changed, self._on_tracing_relation_changed)
         self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
         self.framework.observe(self.on.list_receivers_action, self._on_list_receivers_action)
         self.framework.observe(self.cert_handler.on.cert_changed, self._on_cert_handler_changed)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.tracing_relation_broken, self._on_tracing_relation_broken)
```
PietroPasotti, line 119: I thought we'd agree to keep this in the tracing lib and add a new custom event 
```suggestion
        self.framework.observe(self.tracing.on.broken, self._on_tracing_broken)
```

src/tempo.py
```diff
@@ -104,6 +105,24 @@ def url(self) -> str:
         scheme = "https" if self.tls_ready else "http"
         return f"{scheme}://{self._external_hostname}"
 
+    def get_receiver_url(self, protocol: ReceiverProtocol, ingress: TraefikRouteRequirer):
+        """Return the receiver endpoint URL based on the protocol.
+
+        if ingress is used, return endpoint provided by the ingress instead.
+        """
+        protocol_type = ReceiverProtocolType.get(protocol)
+        is_ingress = ingress.is_ready()
```
PietroPasotti, line 114: nit: 

```suggestion
        has_ingress = ingress.is_ready()
```

or `ingress_is_ready`

src/tempo.py
```diff
@@ -104,6 +105,24 @@ def url(self) -> str:
...
+        receiver_port = self.receiver_ports[protocol]
+
+        url = self.url
+        if is_ingress:
```
PietroPasotti, line 118: the logic here should be cleaned up, maybe with one big if/elif?

PietroPasotti (changes requested):

PietroPasotti (approved):
gj!

lib/charms/tempo_k8s/v1/tracing.py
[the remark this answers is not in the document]
PietroPasotti: I agree with leon on this, the library file should be kept there, and any unit tests on the library itself (not the charm's usage of it) should also be kept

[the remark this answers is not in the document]
PietroPasotti: I mean, it'd be easy enough to recover it if we ever need to backport some fixes, so we're not being super strict in this practice (see traefik's ingress), but still...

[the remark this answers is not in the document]
michaeldmitry: added v1 lib back

PR #127, opened 2024-06-08
S3 support

### LINKED ISSUE #62: Add support for object storage
### Enhancement Proposal

At the moment tempo uses a `local` storage profile: https://github.com/canonical/tempo-k8s-operator/blob/main/src/tempo.py#L122

For a mature traces backend we need to support object storage providers that are available for tempo: https://grafana.com/docs/tempo/latest/setup/operator/object-storage/

mmkay:
## Issue
Add s3 support to tempo charm. Closes #62 

## Solution
Add `s3` interface that can be used by s3-integrator.

## Context
Even though [mimir_config is already using types](https://github.com/canonical/mimir-coordinator-k8s-operator/blob/main/src/mimir_config.py#L104), I haven't switched to typing yet as I expect we'll have more changes with tempo_cluster soon.

## Testing Instructions
In theory the following bundle should work, but it still requires manual steps. Follow the steps added in HACKING.md to deploy minio alongside:

```
bundle: kubernetes
applications:
  minio:
    charm: minio
    channel: latest/edge
    revision: 316
    base: ubuntu@20.04/stable
    resources:
      oci-image: 543
    scale: 1
    options:
      access-key: minio-access-key
      secret-key: minio-secret
    constraints: arch=amd64
    storage:
      minio-data: kubernetes,1,10240M
    trust: true
  s3-integrator:
    charm: s3-integrator
    channel: latest/stable
    revision: 17
    scale: 1
    options:
      bucket: tempo
    constraints: arch=amd64
  tempo:
    charm: local:tempo-k8s-1
    scale: 1
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
relations:
- - tempo:s3
  - s3-integrator:s3-credentials
```

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,3 +1,5 @@
+# Deploying Tempo
+
 To manually test and develop tempo:
 
     charmcraft pack 
@@ -9,3 +11,37 @@ you need to always deploy at least 2 units of tester:
     j deploy ./tester_ubuntu-22.04-amd64.charm tester -n 3 --resource workload=python:slim-buster
     j relate tempo:tracing tester:tracing
 
+# Object storage
+
+In order to test tempo with object storage (using [minio charm](https://github.com/canonical/minio-operator)), 
+you can do the following (replace YOUR_MINIO_ACCESS_KEY and YOUR_MINIO_SECRET with anything you want as long as the
+secret is more than 8 characters):
+
+Deploy minio:
+```
+juju deploy minio --channel edge --trust
+juju config minio access-key=YOUR_MINIO_ACCESS_KEY
+juju config minio secret-key=YOUR_MINIO_SECRET
+```
+
+Deploy mc and set up buckets:
+```
+sudo snap install minio-client --edge --devmode
+kubectl port-forward service/minio -n test-minio 9000:9000
+minio-client config host add local http://localhost:9000 YOUR_MINIO_ACCESS_KEY YOUR_MINIO_SECRET
+minio-client mb local/tempo
+```
+
+Deploy and configure s3-integrator:
+```
+juju deploy s3-integrator
+juju config s3-integrator endpoint=http://MINIO_IP_ADDRESS:9000
+juju config s3-integrator bucket=tempo
+juju run s3-integrator/leader sync-s3-credentials access-key=YOUR_MINIO_ACCESS_KEY secret-key=YOUR_MINIO_SECRET
+```
+
+Deploy tempo:
+```
+juju deploy ./tempo-k8s_ubuntu-22.04-amd64.charm tempo --resource tempo-image=grafana/tempo:2.4.0
+juju integrate tempo s3-integrator
+```
--- a/lib/charms/data_platform_libs/v0/s3.py
+++ b/lib/charms/data_platform_libs/v0/s3.py
@@ -1,0 +1,791 @@
+# Copyright 2023 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+r"""A library for communicating with the S3 credentials providers and consumers.
+
+This library provides the relevant interface code implementing the communication
+specification for fetching, retrieving, triggering, and responding to events related to
+the S3 provider charm and its consumers.
+
+### Provider charm
+
+The provider is implemented in the `s3-provider` charm which is meant to be deployed
+alongside one or more consumer charms. The provider charm is serving the s3 credentials and
+metadata needed to communicate and work with an S3 compatible backend.
+
+Example:
+```python
+
+from charms.data_platform_libs.v0.s3 import CredentialRequestedEvent, S3Provider
+
+
+class ExampleProviderCharm(CharmBase):
+    def __init__(self, *args) -> None:
+        super().__init__(*args)
+        self.s3_provider = S3Provider(self, "s3-credentials")
+
+        self.framework.observe(self.s3_provider.on.credentials_requested,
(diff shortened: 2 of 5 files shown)


src/tempo.py
```diff
@@ -345,6 +338,52 @@ def generate_config(self, receivers: Sequence[ReceiverProtocol]) -> dict:
...
+                    "path": self.wal_path
+                },
+                "pool": {
+                    # number of traces per index record
+                    "max_workers": 400,
+                    "queue_depth": 20000,
+                },
+            }
+        }
+
```
IbraAoad, line 386:
```suggestion
    def _build_storage_config(self, s3_config: Optional[dict] = None):
        trace_config = {
            "wal": {"path": self.wal_path},  # where to store the wal locally
            "pool": {
                "max_workers": 400,  # number of traces per index record
                "queue_depth": 20000,
            },
        }

        if s3_config:
            trace_config.update(
                {
                    "backend": "s3",
                    "s3": {
                        "bucket": s3_config["bucket"],
                        "access_key": s3_config["access-key"],
                        # remove scheme to avoid "Endpoint url cannot have fully qualified paths." on Tempo startup
                        "endpoint": urlparse(s3_config["endpoint"]).netloc,
                        "secret_key": s3_config["secret-key"],
                        "insecure": not s3_config["endpoint"].startswith("https://"),
                    },
                }
            )
        else:
            trace_config.update(
                {
                    "backend": "local",
                    "local": {"path": "/traces"},
                }
            )

        return {"trace": trace_config}
```

src/charm.py
```diff
@@ -111,6 +114,8 @@ def __init__(self, *args):
             self.on.tempo_pebble_custom_notice, self._on_tempo_pebble_custom_notice
         )
         self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.s3_requirer.on.credentials_changed, self._on_s3_changed)
+        self.framework.observe(self.s3_requirer.on.credentials_gone, self._on_s3_changed)
```
IbraAoad, line 118: I understand that this works for monolithic, but maybe we should start thinking of making S3 a requirement if tempo is replicated, please ignore this if this is already in plan for the tempo coordinator/worker

src/charm.py
```diff
@@ -154,6 +159,21 @@ def tls_available(self) -> bool:
             and (self.cert_handler.ca_cert is not None)
         )
 
+    @property
+    def _s3_config(self) -> Optional[dict]:
```
IbraAoad, line 163: maybe also leave a comment or a TODO here explaining why we went with a dict instead of the pydantic types

HACKING.md
```diff
@@ -9,3 +11,37 @@ you need to always deploy at least 2 units of tester:
     j deploy ./tester_ubuntu-22.04-amd64.charm tester -n 3 --resource workload=python:slim-buster
     j relate tempo:tracing tester:tracing
 
+# Object storage
```
IbraAoad, line 14: This is really helpful, we should have drafted smth like this for mimir as well

IbraAoad (changes requested):

src/charm.py
PietroPasotti: that is indeed how it should be. let me check if it's already implemented...

PietroPasotti: nope, doing it now.

src/tempo.py
PietroPasotti: refactored

src/charm.py
PietroPasotti: yeh, don't like that either. Will spend some time to type it all properly

PietroPasotti:
YESSSS

mmkay:
two more things still pending from manual tests, not merging yet

PietroPasotti (approved):
LGTM
note to self: next time, let's keep the formatting changes aside :/

tests/integration/tester-grpc/src/resources/webserver.py
```diff
@@ -168,7 +167,7 @@ def init_overlord():
                     print(f"received {resp} from {subordinate}")
                     try:
                         id_, value_ = resp.text.split("=")
-                    except:
+                    except:  # noqa
```
dimaqq, line 170: Is this change merely because `except:  # noqa` is shorter than `except Exception:`?

Or are you expecting some weird exception like BaseException or forbidden non-exception-ey base?

tests/manual/facades/s3.yaml
```diff
@@ -0,0 +1,8 @@
+endpoint: provide-s3
+app_data: |
+  {
+    "access-key": "mykey",
+    "bucket": "BuckyMcBucket",
+    "endpoint": "http://1.2.3.4:9000",
+    "secret-key": "soverysecret"
+  }
```
dimaqq, line 8: yaml pre-commit could be helpful here

tests/scenario/test_charm.py
```diff
@@ -33,18 +34,42 @@ def test_start(context, base_state):
...
+    # WHEN
+    # the charm receives an ingress(v2) relation-changed requesting an otlp_grpc receiver
+    state = State(leader=True, containers=[tempo], relations=[ingress])
+    context.run(ingress.changed_event, state)
+
+    # THEN
+    # Tempo pushes a new config to the container filesystem
+    fs = tempo.get_filesystem(context)
+    cfg_path = Path(str(fs) + Tempo.config_path)
+    new_config = yaml.safe_load(cfg_path.read_text())
```
dimaqq, line 53: I imagine that this sequence will be repeated, in which case, would you consider:
- either moving it to a fixture
- or providing a convenience method or function?

lib/charms/data_platform_libs/v0/s3.py
```diff
@@ -0,0 +1,791 @@
+# Copyright 2023 Canonical Ltd.
```
dimaqq, line 1: This file is vendored, isn't it?

dimaqq (commented):
Random fly-by review

IbraAoad (approved):

tests/integration/tester-grpc/src/resources/webserver.py
PietroPasotti: no I think that was pure laziness. It's testing code so we didn't mind too much about linting.

tests/scenario/test_charm.py
PietroPasotti: good point, done

PR #136, opened 2024-06-14
instance attribute support for charm_tracing forward-declared definitions

PietroPasotti:
This PR adds support for charm_tracing to grab the endpoint and cert it needs from instance attributes instead of properties and methods.

This is a nicety which in combination with #135 means any charm can enable/disable charm tracing and correctly configure it with/without TLS with a single line of code (plus the decorator).

before:

```
from lib....charm_tracing import charm_tracing

@charm_tracing(tracing_endpoint="foo", server_cert="bar")
class MyCharm(...):
    [...]
    def foo(self):
        # complicated logic to return None if we don't have tracing or we have tracing and tls but no cert yet
        return ... 
    def bar(self):
        return ...
```

after:
```
from lib....charm_tracing import charm_tracing
from lib....tracing import charm_tracing_config

@charm_tracing(tracing_endpoint="foo", server_cert="bar")
class MyCharm(...):
    def __init__(...):
        self.tracing = TracingEndpointRequirer(...)
        self.foo, self.bar = charm_tracing_config(self.tracing, "path/to/cert/")
```

Additionally:
- tuned down logs
- be less forgiving and raise if tracing_endpoint returns rubbish or raises, like we're doing in charm_logging

### DIFF
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -153,6 +153,10 @@ LIBPATCH = 10
 PYDEPS = ["opentelemetry-exporter-otlp-proto-http==1.21.0"]
 
 logger = logging.getLogger("tracing")
+dev_logger = logging.getLogger("tracing-dev")
+
+# set this to 0 if you are debugging/developing this library source
+dev_logger.setLevel(logging.CRITICAL)
 
 tracer: ContextVar[Tracer] = ContextVar("tracer")
 _GetterType = Union[Callable[[CharmBase], Optional[str]], property]
@@ -258,7 +262,7 @@ def _get_tracing_endpoint(
             f"got {tracing_endpoint} instead."
         )
 
-    # logger.debug(f"Setting up span exporter to endpoint: {tracing_endpoint}/v1/traces")
+    dev_logger.debug(f"Setting up span exporter to endpoint: {tracing_endpoint}/v1/traces")
     return f"{tracing_endpoint}/v1/traces"
 
 
@@ -301,7 +305,11 @@ def _setup_root_span_initializer(
         # from the perspective of the charm. Self.unit.name...
 
         original_init(self, framework, *args, **kwargs)
+        # we call this from inside the init context instead of, say, _autoinstrument, because we want it to
+        # be checked on a per-charm-instantiation basis, not on a per-type-declaration one.
         if not is_enabled():
+            # this will only happen during unittesting, hopefully, so it's fine to log a
+            # bit more verbosely
             logger.info("Tracing DISABLED: skipping root span initialization")
             return
 
@@ -373,6 +381,7 @@ def _setup_root_span_initializer(
 
         @contextmanager
         def wrap_event_context(event_name: str):
+            dev_logger.info(f"entering event context: {event_name}")
             # when the framework enters an event context, we create a span.
             with _span("event: " + event_name) as event_context_span:
                 if event_context_span:
@@ -386,6 +395,7 @@ def _setup_root_span_initializer(
 
         @functools.wraps(original_close)
         def wrap_close():
+            dev_logger.info("tearing down tracer and flushing traces")
             span.end()
             opentelemetry.context.detach(span_token)  # type: ignore
             tracer.reset(_tracer_token)
@@ -494,7 +504,7 @@ def _autoinstrument(
     :param extra_types: pass any number of types that you also wish to autoinstrument.
         For example, charm libs, relation endpoint wrappers, workload abstractions, ...
     """
-    # logger.info(f"instrumenting {charm_type}")
+    dev_logger.info(f"instrumenting {charm_type}")
     _setup_root_span_initializer(
         charm_type,
         tracing_endpoint_attr,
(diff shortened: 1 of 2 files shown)


lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -232,8 +244,18 @@ class UntraceableObjectError(TracingError):
...
-def _get_tracing_endpoint(tracing_endpoint_getter, self, charm):
-    if isinstance(tracing_endpoint_getter, property):
+class TLSError(TracingError):
+    """Raised when the tracing endpoint is https but we don't have a cert yet."""
+
+
+def _get_tracing_endpoint(
+    tracing_endpoint_getter: Union[_GetterType, _InstanceAttribute],
+    self: ops.CharmBase,
+    charm: Type[ops.CharmBase],
```
ca-scribner, line 254: since `self` is by convention usually meaning *this object* in python, maybe name it differently?  `charm_instance` and `charm_class` could work.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -232,8 +244,18 @@ class UntraceableObjectError(TracingError):
...
+):
+    if isinstance(tracing_endpoint_getter, _InstanceAttribute):
+        tracing_endpoint = getattr(self, tracing_endpoint_getter)
+    elif isinstance(tracing_endpoint_getter, property):
```
ca-scribner, line 258: (for some reason I thought class attributes were also a possibility, so I wrote this with them in mind too.  But iiuc they're not, and you can ignore that part :) )

Can you get rid of `_InstanceAttribute` by doing a duck-typing approach?  Here's a demo of what I'm thinking:

```python
def get_val(obj, attr):
    """Returns the value of an attribute in an object that could be a method, property, instance variable, or class variable."""
    value_getter = getattr(obj, attr)
    try:
        value = value_getter()
        print("Callable - this was a method!")
    except TypeError:
        print("Not callable, so its a value or property that rendered itself to a value")
        value = value_getter
    return value

class Test:
    class_attribute = "class attribute"
    class_attribute_we_will_mask = "class attribute not yet masked"
    class_attribute_we_will_change_in_class = "class attribute not yet changed in class"

    def __init__(self, masked_class_attribute=None):
        self.instance_attribute = "instance attribute"
        if masked_class_attribute:
            self.class_attribute_we_will_mask = masked_class_attribute

    @property
    def my_property(self):
        return "my property"

    def my_method(self):
        return "my method"

t_unmasked = Test()
t_masked = Test("class attribute masked")

def print_results(x):
    print("___________________")
    print(f"{get_val(x, 'my_property')=}")
    print(f"{get_val(x, 'class_attribute')=}")
    print(f"{get_val(x, 'class_attribute_we_will_mask')=}")
    print(f"{get_val(x, 'instance_attribute')=}")
    print(f"{get_val(x, 'my_method')=}")

print("t_unmasked")
print_results(t_unmasked)

# OUTPUT:
# t_unmasked
# ___________________
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'my_property')='my property'
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'class_attribute')='class attribute'
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'class_attribute_we_will_mask')='class attribute not yet masked'
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'instance_attribute')='instance attribute'
# Callable - this was a method!
# get_val(x, 'my_method')='my method'

print("t_masked")
print_results(t_masked)

# OUTPUT
# t_masked
# ___________________
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'my_property')='my property'
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'class_attribute')='class attribute'
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'class_attribute_we_will_mask')='class attribute masked'
# Not callable, so its a value or property that rendered itself to a value
# get_val(x, 'instance_attribute')='instance attribute'
# Callable - this was a method!
# get_val(x, 'my_method')='my method'


```

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: I'm keeping it in the overridden charm init though, it makes it easier to reason about what's going on if you think from the perspective of the charm instance

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: this helps! I just have to delay the getattr call to simplify the current approach quite a lot.

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -500,12 +515,12 @@ def trace_type(cls: _T) -> _T:
     It assumes that this class is only instantiated after a charm type decorated with `@trace_charm`
     has been instantiated.
     """
-    logger.info(f"instrumenting {cls}")
+    # logger.info(f"instrumenting {cls}")
```
mmkay, line 518: Prtetty useful for debugging, but maybe we want to remove the commented out lines before the merge?

mmkay (approved):
Muuuuch nicer, thanks @ca-scribner!

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: yeh I was considering to keep them there because they're very useful in development. should we add a separate logger for them perhaps, and manually enable that when needed?

mmkay: Sounds like a great idea :)

PR #139, opened 2024-06-18
add service name -charm prefix

### LINKED ISSUE #88: give different default service name for charm_tracing
### Bug Description

charm traces' service defaults to <application name>, but that can be confusing.
E.g. if you deploy `grafana-k8s` as `grafana`, the service name will be 'grafana', which is also the service name of grafana's workload traces. 

Result is two services getting mushed into one. 

Consider adding a `-charm` suffix to the default service name, so that charm traces appear as `grafana-charm`.

### To Reproduce

n.a

### Environment

n.a

### Relevant log output

```shell
n.a
```

### Additional context

_No response_

PietroPasotti:
Fixes #88 

# Issue:

If you deploy `grafana-k8s` as `grafana`, the service name for charm traces was 'grafana'. The traces generated by the grafana workload also had the same service name, resulting in confusion.

# Solution:

Service name for the charm tracer will now default to `<app name>-charm` so a conflict with the workload service name is less likely.

### DIFF
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -294,8 +294,8 @@ def _setup_root_span_initializer(
         # self.handle = Handle(None, self.handle_kind, None)
 
         original_event_context = framework._event_context
-
-        _service_name = service_name or self.app.name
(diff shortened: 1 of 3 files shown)


lucabello (approved):
Make sure CI is green before merging :)

PR #140, opened 2024-06-18
changed root span name to be more descriptive

### LINKED ISSUE #111: Add more identifying information to the trace query result table
### Enhancement Proposal

Currently, the following is the layout of the trace query result table:
image

The proposal is to add more identifying information to this table so that the user has enough context to pick the trace that they are interested in. Off the top of my head, the following information may be useful:
1. The unit of the trace (which will inherently imply the application of the trace)
2. The hook of the trace (e.g. `update-status` or `a-relation-created`)

Additionally, providing filtering mechanisms may be very useful. For instance, if we want to see what happened during relation formation, we can query `a-relation-created` event traces and just look at traces for this hook (across various units and applications)

PietroPasotti:
Fixes #111 

## Issue
right now all charm traces that show up in grafana have the same title: `charm exec`. 
Not very descriptive.

## Solution
Now it looks like `<unit-name>: <event> event`, for example `loki/1: foo-relation-changed event`

### DIFF
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -146,7 +146,7 @@ LIBAPI = 1
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 8
+LIBPATCH = 9
 
 PYDEPS = ["opentelemetry-exporter-otlp-proto-http==1.21.0"]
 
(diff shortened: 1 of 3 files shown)


Abuelodelanada (approved):

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -341,16 +342,18 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
         _tracer = get_tracer(_service_name)  # type: ignore
         _tracer_token = tracer.set(_tracer)
 
-        dispatch_path = os.getenv("JUJU_DISPATCH_PATH", "")
+        dispatch_path = os.getenv("JUJU_DISPATCH_PATH", "")  # something like hooks/install
+        event_name = dispatch_path.split("/")[1] if "/" in dispatch_path else dispatch_path
+        root_span_name = f"{unit_name}: {event_name} event"
```
mmkay, line 347: Do we need the `event` word? it will be in every event so I'm not sure if that makes sense to have it there.

michaeldmitry (approved):

michaeldmitry:
Please increment patch

PR #144, opened 2024-06-24
Better docs server cert

### LINKED ISSUE #90: Please improve documentation on the `server_cert` parameter inside of `trace_charm`
### Enhancement Proposal

From my understanding the charm is sending its telemetry to Tempo (hopefully via https, but even this is not obvious). If that is the case, then why does the charm need a certificate? Is there a server running on the charm? Is it Tempo's certificate? If that's the case then how would the charm know about it?

The doc let us believe that not setting this leads to sending traces over an insecure connection. But why, how? It's the server that needs a TLS cert, not the client. Is there some mTLS going on?

In any case, this is not quite clear at the moment and better documentation on this topic would be helpful to charm developers.

PietroPasotti:
Fixes #90


Abuelodelanada (approved):

ca-scribner (commented):
Everything about this PR lgtm, but does it close #90?  I don't see the clarifications that were requested

mmkay (approved):

michaeldmitry (approved):

lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -9,22 +9,57 @@
...
+```python
+# import the necessary charm libs
+from charms.tempo_k8s.v2.tracing import TracingEndpointRequirer, charm_tracing_config
+from charms.tempo_k8s.v1.charm_tracing import charm_tracing
+
+# decorate your charm class with charm_tracing:
+@charm_tracing(
+    # forward-declare the instance attributes that the instrumentor will look up to obtain the
+    # tempo endpoint and server certificate
+    tracing_endpoint="charm_tracing_endpoint",
```
mmkay, line 29: Should it be tracing_endpoint (without charm_) as the properties are named below in __init__? Same with server_cert.

PietroPasotti:
> Everything about this PR lgtm, but does it close #90? I don't see the clarifications that were requested

lines 33-38
lines 73-79

PR #149, opened 2024-07-02
removed excess receiver protocols from enum

PietroPasotti:
## Issue
This helps address issue #https://github.com/canonical/tempo-coordinator-k8s-operator/issues/8

### DIFF
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -121,7 +121,6 @@ ReceiverProtocol = Literal[
     "zipkin",
     "otlp_grpc",
     "otlp_http",
-    "opencensus",
     "jaeger_grpc",
     "jaeger_thrift_http",
 ]
@@ -145,7 +144,6 @@ receiver_protocol_to_transport_protocol: Dict[ReceiverProtocol, TransportProtoco
     "zipkin": TransportProtocolType.http,
     "otlp_grpc": TransportProtocolType.grpc,
     "otlp_http": TransportProtocolType.http,
-    "opencensus": TransportProtocolType.grpc,
     "jaeger_thrift_http": TransportProtocolType.http,
     "jaeger_grpc": TransportProtocolType.grpc,
 }
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -124,8 +124,6 @@ ReceiverProtocol = Literal[
     "opencensus",
     "jaeger_grpc",
     "jaeger_thrift_http",
-    "jaeger_thrift_compact",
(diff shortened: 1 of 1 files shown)


lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -118,10 +118,6 @@ def __init__(self, *args):
 
 ReceiverProtocol = Literal[
     "zipkin",
-    "kafka",
-    "opencensus",
-    "tempo_http",
-    "tempo_grpc",
     "otlp_grpc",
```
mmkay, line 121: The protocols aren't needed in `receiver_protocol_to_transport_protocol` either:

https://github.com/canonical/tempo-k8s-operator/pull/149/files#diff-b47b4334954f653c5e1d7a8831211b30280d18fa54a07681f9108cec9ce34923R140 

(where I somehow can't leave a comment)

PietroPasotti: ehm, funny enough, it seems that Tempo is supporting jaeger thrift

PietroPasotti: can go either way with it now. Do we add it to the supported protocols or strip it completely?

PietroPasotti: ah but wait, we don't really care. This is monolithic tempo. IIRC we've removed it already in the distributed tempo?

PietroPasotti: nope, still there.

mmkay: It's in an uncertain state at the moment: https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/tempo.py#L53 still has it. I can remove it together with nginx PR if you think it's the way to go. 
@michaeldmitry, any opinions here maybe?

michaeldmitry: My understanding is that we're going to drop Jaeger support at some point? https://www.jaegertracing.io/docs/1.58/apis/#opentelemetry-protocol-stable
That's why I didn't expose it as a config option, but looks like I missed it in https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/tempo.py#L53.

mmkay (approved):

PietroPasotti:
maybe obsoleted by https://github.com/canonical/observability/pull/190

lib/charms/tempo_k8s/v2/tracing.py
michaeldmitry: Continuing in https://github.com/canonical/tempo-coordinator-k8s-operator/issues/8

michaeldmitry:
Added supported protocols list as per ADR https://github.com/canonical/observability/pull/196

PR #160, opened 2024-08-07
Fix resume error

PietroPasotti:
On sleep/resume, the tempo charm goes into error because it receives a custom notice event for a service that doesn't exist in pebble (yet)?
Unclear what's causing this, but it's a pain.
Workaround and test included.

### DIFF
--- a/tests/scenario/test_charm.py
+++ b/tests/scenario/test_charm.py
@@ -1,6 +1,6 @@
 import os
 from pathlib import Path
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 import yaml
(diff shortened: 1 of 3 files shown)


mmkay (approved):

tests/scenario/test_charm.py
```diff
@@ -179,3 +184,22 @@ def test_tracing_storage_is_configured_to_s3_if_s3_relation_filled(
...
+    container, tempo = _tempo_mock_with_initial_config(tmp_path, tempo_ready_svc_exists=False)
+
+    state = State(leader=True, containers=[tempo])
+
+    # WHEN we receive a custom-notice event
+    with caplog.at_level("DEBUG"):
+        os.environ["SCENARIO_SKIP_CONSISTENCY_CHECKS"] = "1"
+        # scenario doesn't play nice in this very edge case
+        context.run(_BoundNotice(Notice(Tempo.tempo_ready_notice_key), tempo).event, state)
+        del os.environ["SCENARIO_SKIP_CONSISTENCY_CHECKS"]
```
dimaqq, line 200: Minor: consider https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setenv instead

PietroPasotti: uh, cool, didn't know about that

PR #161, opened 2024-08-13
solution and test for tracer multi-import issue

### LINKED ISSUE #60: Tracer contextvar is sometimes empty
After the switch to the new dependency [otlp_http] it seems there's been a change in the way `tracer` behaves in the context here. I started getting scenario test failures on `test_base_tracer_endpoint_event` so I started debugging and noticed that `_get_tracer()` returns None in the context of the span in the function instrumented by `@trace`. After going through the context I noticed that there is a `tracer` in the context but with a different reference, so I added this fallback retrieval.

To test it doesn't only happen on my local environment I've tried removing the fallback on a test draft PR and it looks the same behavior happens there: https://github.com/canonical/tempo-k8s-operator/actions/runs/7412918066/job/20170743398?pr=58

Branch diff: https://github.com/canonical/tempo-k8s-operator/compare/switch-to-otlp-http...temp/otlp-http-without-get-context

As fallback is triggered only in case `tracer.get()` doesn't find anything I felt it's okay to add it here in that way. I must admit though that it's dealing with a symptom, not addressing the cause I couldn't really pinpoint.

Do you think we should open a bug to investigate that further?

_Originally posted by @mmkay in https://github.com/canonical/tempo-k8s-operator/pull/55#discussion_r1442127616_

PietroPasotti:
## Issue
Fixes #60 

## Solution
Added a warning if we're detecting a double import situation.

## Context
https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap

## Testing Instructions
Added a test verifying the warning behaviour.

### DIFF
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -227,11 +227,24 @@ _remove_stale_otel_sdk_packages()
 import functools
 import inspect
 import logging
-import opentelemetry
-import ops
 import os
 from contextlib import contextmanager
 from contextvars import Context, ContextVar, copy_context
+from pathlib import Path
+from typing import (
+    Any,
+    Callable,
+    Generator,
+    Optional,
+    Sequence,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+)
+
+import opentelemetry
+import ops
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
 from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace import Span, TracerProvider
@@ -247,18 +260,6 @@ from opentelemetry.trace import (
 from opentelemetry.trace import get_current_span as otlp_get_current_span
 from ops.charm import CharmBase
 from ops.framework import Framework
-from pathlib import Path
-from typing import (
-    Any,
-    Callable,
-    Generator,
-    Optional,
-    Sequence,
-    Type,
-    TypeVar,
-    Union,
-    cast,
-)
 
 # The unique Charmhub library identifier, never change it
 LIBID = "cb1705dcd1a14ca09b2e60187d1215c7"
@@ -269,7 +270,7 @@ LIBAPI = 1
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 
-LIBPATCH = 14
+LIBPATCH = 15
 
 PYDEPS = ["opentelemetry-exporter-otlp-proto-http==1.21.0"]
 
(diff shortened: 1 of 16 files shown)


lib/charms/tempo_k8s/v1/charm_tracing.py
```diff
@@ -333,9 +331,17 @@ def _get_tracer() -> Optional[Tracer]:
...
+        # from different paths (typically charms.tempo_k8s... and lib.charms.tempo_k8s...)
         try:
             ctx: Context = copy_context()
             if context_tracer := _get_tracer_from_context(ctx):
+                logger.warning(
+                    "Tracer not found in `tracer` context var. "
+                    "Verify that you're importing all `charm_tracing` symbols from the same module path."
+                    "For more info: https://python-notes.curiousefficiency.org/en/latest/python"
+                    "_concepts/import_traps.html#the-double-import-trap"
+                )
```
ca-scribner, line 344: Can we add to this message something concrete for a charmer?  I can see someone who isn't really versed in python reading this reference and still thinking "but what do I do to fix this?".  Maybe there's a more specific example of how they could have messed it up in charm code and how to fix?  Could put those details in a gh issue and link it here

tests/scenario/test_a_charm_tracer_multi_import_warning.py
```diff
@@ -0,0 +1,67 @@
...
+# imports from previous tests will pollute the sys.modules and cause this test to fail.
+# I know this is horrible but yea, couldn't find a better way to fix the issue. Tried:
+# - delete from sys.modules all modules containing 'charms.tempo_k8s'
+# - delete from sys.modules all modules containing 'otlp_http'
+
+
+from unittest.mock import patch
+
+# this test file is intentionally quite broken, don't modify the imports
+# import autoinstrument and get_current_span from charms.[...]
```
ca-scribner, line 11: No longer used?

```suggestion
# import autoinstrument from charms.[...]
```

tests/scenario/test_a_charm_tracer_multi_import_warning.py
```diff
@@ -0,0 +1,67 @@
+# WARNING keep this test module on top of the scenario tests, else the
+# imports from previous tests will pollute the sys.modules and cause this test to fail.
```
ca-scribner, line 2: by "on top of" do you mean separate?  Like if we merged these tests into the existing other scenario test files, we'd have a conflict?  If yes, what about rewording:

```suggestion
# WARNING keep these tests separate to other scenario tests , else the
# imports from previous tests will pollute the sys.modules and cause this test to fail.
```

ca-scribner (approved):
afaict this makes sense, but I am not so well versed so a second opinion might be warranted.

tests/scenario/test_a_charm_tracer_multi_import_warning.py
PietroPasotti: no I mean that the file name starts with an "a" so that it's lexicographically first and is executed before any other scenario test. Will rephrase.

lib/charms/tempo_k8s/v1/charm_tracing.py
PietroPasotti: I think it pays off in this case to make the warning more long and explicit.

PR #170, opened 2024-09-17
Add using `is_ready` safeguard to `get_endpoint` docstring

### LINKED ISSUE #169: Non-leader unit charm errors out on `tracing-relation-xx` event when integrated with Tempo over `tracing`
### Bug Description

Occasionally, in CI, When a charm X is deployed with multiple units and integrated with Tempo over `tracing`, the non-leader unit(s) could error out during `tracing-relation-xx` event, as its trying to access the charm's app databag to get the receivers requested.

This probably happens due to [this line](https://github.com/canonical/tempo-k8s-operator/blob/main/lib/charms/tempo_k8s/v2/tracing.py#L912) when Tempo hasn't yet published the endpoints requested by the requirer charm X, but the charm X is trying to get the endpoint and when it doesn't find them in Tempo's app databag, it tries to load its own app databag to check the receivers it has requested.

### To Reproduce

Deploy a multi-unit charm and integrate it with Tempo over `tracing`.
It **occasionally** happens. 

### Environment

microk8s: MicroK8s v1.28.13 revision 7206
juju: 3.4.5-genericlinux-amd64

### Relevant log output

```shell
Check additional context
```

### Additional context

Screenshot from 2024-09-17 09-35-11
image

michaeldmitry:
## Issue
Fixes #169 

## Solution
Add documentation to use the `is_ready` safeguard before `get_endpoint` to prevent permission denied exceptions due to non-leader units trying to access local app databags.

### DIFF
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -906,7 +906,7 @@ class TracingEndpointRequirer(Object):
 
         It could happen that this function gets called before the provider publishes the endpoints.
         In such a scenario, if a non-leader unit calls this function, a permission denied exception will be raised due to
-        unauthorized access to the local app databag. To prevent this, this function needs to be guarded by the `is_ready` check.
+        restricted access. To prevent this, this function needs to be guarded by the `is_ready` check.
 
         Raises:
         ProtocolNotRequestedError:
--- a/lib/charms/tempo_k8s/v2/tracing.py
+++ b/lib/charms/tempo_k8s/v2/tracing.py
@@ -107,7 +107,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 10
 
 PYDEPS = ["pydantic"]
 
@@ -902,7 +902,16 @@ class TracingEndpointRequirer(Object):
     def get_endpoint(
         self, protocol: ReceiverProtocol, relation: Optional[Relation] = None
     ) -> Optional[str]:
-        """Receiver endpoint for the given protocol."""
+        """Receiver endpoint for the given protocol.
+
+        It could happen that this function gets called before the provider publishes the endpoints.
+        In such a scenario, if a non-leader unit calls this function, a permission denied exception will be raised due to
+        unauthorized access to the local app databag. To prevent this, this function needs to be guarded by the `is_ready` check.
+
+        Raises:
+        ProtocolNotRequestedError:
+            If the charm unit is the leader unit and attempts to obtain an endpoint for a protocol it did not request.
+        """
         endpoint = self._get_endpoint(relation or self._relation, protocol=protocol)
         if not endpoint:
             requested_protocols = set()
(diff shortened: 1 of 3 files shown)


mmkay (approved):

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -902,9 +902,19 @@ def _get_endpoint(
...
+
+        Raises:
+        ProtocolNotRequestedError:
+            If the charm unit is the leader unit and attempts to obtain an endpoint for a protocol it did not request.
+        """
         endpoint = self._get_endpoint(relation or self._relation, protocol=protocol)
         if not endpoint:
+            # non-leader units cannot access local app databag
+            if not self._charm.unit.is_leader():
+                # if its a racing condition, we can return None until the provider publishes the endpoint in a later event.
```
PietroPasotti, line 915:
```suggestion
                # if its a race condition, we can return None until the provider publishes the endpoint in a later event.
```
Also it's not clear to me what is racing with what?

lib/charms/tempo_k8s/v2/tracing.py
```diff
@@ -902,9 +902,19 @@ def _get_endpoint(
...
+        """Receiver endpoint for the given protocol.
+
+        Raises:
+        ProtocolNotRequestedError:
+            If the charm unit is the leader unit and attempts to obtain an endpoint for a protocol it did not request.
+        """
         endpoint = self._get_endpoint(relation or self._relation, protocol=protocol)
         if not endpoint:
+            # non-leader units cannot access local app databag
+            if not self._charm.unit.is_leader():
```
PietroPasotti, line 914: It's always a of a trade-off, but I wonder if in this case it wouldn't be better to raise an exception if we are not the leader and force the caller to guard instead.
Reason is that the caller should be aware that only leaders can see the endpoints, and now if this call returns None, we don't know if it's because there is no endpoint (yet) or because we're not allowed to see it.

lib/charms/tempo_k8s/v2/tracing.py
michaeldmitry: if the requirer tries to get the endpoint before the provider publishes it

lib/charms/tempo_k8s/v2/tracing.py
michaeldmitry: hmmm good point
but, will it make a difference to know the reason why we got None? If there's an issue with the protocol sent, you'll still get notified, as the leader will crash. 

In the case of the non-leader, having None will atleast not break just because of a race condition. Another reason for me is that its not quite easy to catch the issue in terms of how often it'd happen. 

Its 50 50 for me :D

lib/charms/tempo_k8s/v2/tracing.py
PietroPasotti: ah, ok, I don't usually think of those as race conditions but "just juju", but I guess technically they are :)

lib/charms/tempo_k8s/v2/tracing.py
PietroPasotti: Well presumably the user will unittest this, and if you don't set leader, you'll get an error as soon as you call get_endpoint, and you'll know you need a guard. Which by the way, is what happens already, only now you would get a better error.
So I think it's more in line with the current behaviour to keep raising if you're calling this without leadership

michaeldmitry: I'll keep the existing behavior, but will include more info about this in the docstring of the function

PR #172, opened 2024-09-25
disable tracing instead of raising if cert not provided but endpoint is https

PietroPasotti:
problem: 
relate tempo with self-signed certs before you relate traefik to self-signed certs, and traefik will error out before it has any chance to process certificates relation event.

This will also make charm tracing behave more conservatively and avoid at all costs to put the host charm in error state (idea which we previously considered and rejected)


mmkay (approved):
A bit of tests might be beneficial, but the changes look good.

PietroPasotti:
included in https://github.com/canonical/tempo-coordinator-k8s-operator/pull/51

